PR #10, opened 2018-05-24
Automatic dependency update for rfc5424-logging-handler to version 1.1.2

fridex:
Dependency rfc5424-logging-handler was used in version 1.1.0, but the current latest version is 1.1.2.


goern:
@fridex dont forget to write down a feature and provide tests :P

PR #32, opened 2018-07-02
Introduce a function for getting service account token

fridex:
Access to OpenShift/Kubernetes token is done in various services and pods. It 
would be better to utilize this functionality undet the thoth-common package 
not to replicate code in the code base.


goern:
@fridex What about a commit description? Maybe we can extract release notes out of it... some day...

fridex:
@goern updated :+1:

PR #46, opened 2018-07-26
Allow completely suppressing logs

fridex:


fridex:
ping @sushmithaaredhatdev

sushmithaaredhatdev:
This change looks good tried replicating it locally, it works fine.

PR #80, opened 2018-09-03
Fix default TLS verification behavior

fridex:


fridex:
Meh...

goern:
it's running for just 12mins... give it a little more time...

fridex:
> it's running for just 12mins... give it a little more time...

I wanted a review :)

PR #88, opened 2018-09-05
Let's reuse adviser env var names

fridex:


fridex:
Merging to continue with work.

PR #101, opened 2018-09-17
Propagate empty string if runtime env is not set

fridex:


fridex:
This was observed on multiple places, thus replacing with more generic way introduced in https://github.com/thoth-station/common/pull/103

PR #108, opened 2018-09-25
Add Sentry support

fridex:
Depends-On: https://github.com/thoth-station/core/pull/150


fridex:
I was not able to update lock (some local issues), let's use Kebechet for that.

thoth/common/logging.py
```diff
@@ -85,9 +88,16 @@ def init_logging(logging_configuration: dict = None) -> None:
 
     _init_log_levels(logging_configuration)
 
+    if _SENTRY_DSN:
```
goern: any docs on this? what and why to do?

Pipfile
```diff
@@ -7,6 +7,7 @@ name = "pypi"
 "rfc5424-logging-handler" = "*"
 daiquiri = "*"
 requests = "*"
+raven = "*"
```
goern: why raven? it reads like "Raven is the legacy Python client for Sentry (getsentry.com) — replaced by sentry-python https://sentry.io"

fridex: Ah, they are fast. That is the way I did it in the previous project :)

thoth/common/logging.py
fridex: In the core repo.

PR #132, opened 2018-10-04
Introduce methods for running dependency monkey

fridex:


fridex:
Re: jobs vs pods - we will turn the switch at the same time for all analyzers once there will be adjusted logic in the implementation (nice new issue for newcomers). Let's stick with the current pod scheduling for now.

PR #133, opened 2018-10-04
Make reusable methods public

fridex:


fridex:
... so that they can be reused by Amun

fridex:
I will merge this to continue with work.

PR #138, opened 2018-10-05
Add routines for jobs handling

fridex:


fridex:
Done in https://github.com/thoth-station/common/pull/138

PR #143, opened 2018-10-09
Fix undefined name error

fridex:


fridex:
Merging as hotfix, copy-pasta error.

thoth/common/openshift.py
```diff
@@ -189,15 +189,15 @@ def get_build_log(self, build_id: str, namespace: str) -> str:
         """Get log of a build in the given namespace."""
         # TODO: rewrite to OpenShift rest client once it will support it.
         endpoint = "{}/apis/build.openshift.io/v1/namespaces/{}/builds/{}/log".format(
-            _OPENSHIFT.openshift_api_url,
```
goern: could pylint catch that?

fridex: In some cases yes.

fridex: (most of the cases in normal code style)

goern: but this is not a code styling problem, its a matter of the semantics, isnt it?

fridex: This can be determined statically in this case, it looks like the linter is not configured correctly (also, there is another issue in this file - explicitly deriving from object, that should be uncovered by pylint - convention). We will need to revisit configuration (install all the deps so this static checks are performed on dependencies and run the linter correctly)

Issue #164, opened 2018-10-18
provide get_all_jobs(label) OpenShift function


goern:
**Is your feature request related to a problem? Please describe.**
I would like to get all jobs selected by a provided label. all information should be returned as a dict

**Describe the solution you'd like**
...

**Describe alternatives you've considered**
none.

**Additional context**
This is used for metrics.

PR #167, opened 2018-10-19
added get_jobs(), it will be used for thoth-metrics

goern:
This implements #164


goern:
as github does not respect 'implements' this PR also closes #164

PR #187, opened 2018-10-30
Make all datetimes timezone aware

fridex:


fridex:
A hotfix..

goern:
why is everything in UTC not good enough?

fridex:
> why is everything in UTC not good enough?

This query [1] used for checking if the given document was already synced fails due to timezone issues - when computing timestamp, there are issues with timezone not being set.

[1] https://github.com/thoth-station/storages/blob/master/thoth/storages/graph/janusgraph.py#L386

fridex:
@goern I will merge this to check it in the cluster

PR #205, opened 2018-11-14
Use api version from template

fridex:


fridex:
recheck

goern:
a `regate` would have done the job :)

PR #277, opened 2019-01-16
Fix issues when template does not request any resources

fridex:


fridex:
We have chicken egg problem here - inspection jobs have requests as parameters, but we process templates on the actual job creation.

PR #284, opened 2019-01-18
Label workload to allow type specific queries

fridex:


fridex:
For now, later we can introduce CRDs for this purpose.

PR #319, opened 2019-02-14
A temporary workaround for inspection jobs not to be stuck on builds

fridex:


fridex:
Maybe this approach is not good as order of events is not preserved - we can wait for build that has not been scheduled...

PR #357, opened 2019-04-05
refactored retrieval of template

goern:
refactored retrieval of template to OpenShift._get_template(label_selector)

Signed-off-by: Christoph Görn <goern@redhat.com>

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -384,7 +384,7 @@ class OpenShift:
         for key, value in status[state].items():
             if key == "containerID":
                 value = (
-                    value[len("docker://") :]
+                    value[len("docker://"):]
                     if value.startswith("docker://")
                     else value
                 )
@@ -1254,7 +1254,6 @@ class OpenShift:
 
         return self._get_template("template=provenance-checker")
 
-
     def schedule_graph_sync(
         self, document_id: str, namespace: str, *, template_name: str = None
     ):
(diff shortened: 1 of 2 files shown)


goern:
do we have to roll a new release now?

thoth/common/openshift.py
```diff
@@ -382,7 +384,7 @@ def _status_report(status):
         for key, value in status[state].items():
             if key == "containerID":
                 value = (
-                    value[len("docker://"):]
+                    value[len("docker://") :]
```
fridex: Coala will complain here.

thoth/common/openshift.py
```diff
@@ -472,6 +474,19 @@ def get_jobs(self, label_selector: str, namespace: str = None) -> dict:
 
         return response
 
+    def _get_template(self, _label_selector: str) -> dict:
+        """Get template from infra namespace, use label_selector to identify which template to get."""
+        response = self.ocp_client.resources.get(
+            api_version="template.openshift.io/v1", kind="Template"
+        ).get(namespace=self.infra_namespace, label_selector=_label_selector)
+        _LOGGER.debug(
+            "OpenShift response for getting provenance-checker template: %r",
```
fridex: This one is very specific for just provenance-checker, but the function used for other templates as well. What about logging just label selector which should be descriptive enough?

thoth/common/openshift.py
```diff
@@ -687,9 +682,10 @@ def get_solver_names(self) -> list:
                 "Infra namespace is required in order to list solvers"
             )
 
-        response = self.ocp_client.resources.get(api_version="v1", kind="Template").get(
-            namespace=self.infra_namespace, label_selector="template=solver"
-        )
+        # TODO can we refactor this to use _get_template() ?
```
fridex: Maybe introduce `_get_templates()` instead and return template listing. Additional parameter (something like, `sanitize_only_one=True`) would check if there is just one. A check for != 0 would be performed always. I don't insist on this change as it's internal logic which would not introduce tech-dept elsewhere , just suggestion to next implementation.

fridex:
> do we have to roll a new release now?

Yes, we will need to do that.

PR #395, opened 2019-05-22
minor fix for correct namespace

harshad16:
- configmap were being created in backend instead of middletier.


thoth/common/openshift.py
```diff
@@ -816,7 +816,7 @@ def schedule_package_extract(
         job_id: str = None,
     ) -> str:
         """Schedule the given job run, the scheduled job is handled by workload operator based resources available."""
-        if not self.backend_namespace:
+        if not self.middletier_namespace:
             raise ConfigurationError(
                 "Unable to schedule package extract job without backend namespace being set"
```
fridex: >  "Unable to schedule package extract job without **backend** namespace being set"

PR #443, opened 2019-07-17
schedule_package_analyzer

bissenbay:
Created a scheduler for package-analyzer

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -980,6 +980,82 @@ class OpenShift:
 
         return self._get_template("template=package-extract")
 
+    def schedule_package_analyzer(
+        self,
+        package_name: str,
+        package_version: str,
+        index_url: str,
+        *,
+        output: str,
+        debug: bool = False,
+        job_id: str = None,
+    ) -> str:
+        """Schedule the given job run, the scheduled job is handled by workload operator based resources available."""
+        if not self.middletier_namespace:
+            raise ConfigurationError(
+                "Unable to schedule package analyzer job without middletier namespace being set"
+            )
+
+        job_id = job_id or self._generate_id("package-analyzer")
+        parameters = locals()
+        parameters.pop("self", None)
+        return self._schedule_workload(
+            run_method_name=self.run_package_analyzer.__name__,
+            run_method_parameters=parameters,
+            template_method_name=self.get_package_analyzer_template.__name__,
+            job_id=job_id,
+            namespace=self.middletier_namespace,
+            labels={"component": "package-analyzer"},
+        )
+
+    def run_package_analyzer(
+        self,
+        package_name: str,
+        package_version: str,
+        index_url: str,
+        *,
+        output: str,
+        debug: bool = False,
+        job_id: str = None,
+        template: dict = None,
+    ) -> str:
+        """Run package-analyzer to gather digests of packages and files present inside packages."""
+        if not self.middletier_namespace:
+            raise ConfigurationError(
+                "Running package-analyzer requires middletier namespace to be specified"
+            )
+
+        template = template or self.get_package_analyzer_template()
+
+        self.set_template_parameters(
+            template,
+            THOTH_PACKAGE_ANALYZER_PACKAGE_NAME=package_name,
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -980,6 +980,80 @@ def get_package_extract_template(self):
 
         return self._get_template("template=package-extract")
 
+    def schedule_package_analyzer(
+        self,
+        package_name: str,
+        package_version: str,
+        index_url: str,
+        output: str,
```
bissenbay: Package-extractor had additional parameters such as *, environment_type, origin. Do I need them here as well?

thoth/common/openshift.py
```diff
@@ -980,6 +980,80 @@ def get_package_extract_template(self):
 
         return self._get_template("template=package-extract")
 
+    def schedule_package_analyzer(
+        self,
+        package_name: str,
+        package_version: str,
+        index_url: str,
```
bissenbay: Package-extractor had additional parameters such as *, environment_type, origin. Do I need them here as well?

thoth/common/openshift.py
```diff
@@ -980,6 +980,80 @@ def get_package_extract_template(self):
...
+            namespace=self.middletier_namespace,
+            labels={"component": "package-analyzer"},
+        )
+
+    def run_package_analyzer(
+        self,
+        package_name: str,
+        package_version: str,
+        index_url: str,
+        output: str,
```
bissenbay: Should I place * parameter to make sure that there might some extra parameters coming? If yes, then where is it good to place?

thoth/common/openshift.py
fridex: What would these parameters do?

thoth/common/openshift.py
```diff
@@ -980,6 +980,80 @@ def get_package_extract_template(self):
...
+        debug: bool = False,
+        job_id: str = None,
+    ) -> str:
+        """Schedule the given job run, the scheduled job is handled by workload operator based resources available."""
+        if not self.middletier_namespace:
+            raise ConfigurationError(
+                "Unable to schedule package extract job without middletier namespace being set"
```
```diff
-                "Unable to schedule package extract job without middletier namespace being set"
+                "Unable to schedule package analyzer job without middletier namespace being set"
```

thoth/common/openshift.py
```diff
@@ -980,6 +980,80 @@ def get_package_extract_template(self):
...
+            )
+
+        job_id = job_id or self._generate_id("package-analyzer")
+        parameters = locals()
```
fridex: Won't be `job_id` repeated in the run method? (move assignemnt to `parameters` before assignment of to `job_id` environment variable)

thoth/common/openshift.py
fridex: * is used to delimit keywords arguments from positional arguments. You can place it after `index_url`.

thoth/common/openshift.py
bissenbay: @fridex I definitely can move ```parameters``` before ```job_id```. However, I just took a look how other schedulers are written and it seems like all of them have ```job_id``` above ```parameters```. WDYT?

fridex: Agh, right - you are right. The reference to `job_id` is always present. Keep it as is. Nice catch! :+1:

bissenbay: Thanks a lot @fridex !

thoth/common/openshift.py
```diff
@@ -980,6 +980,82 @@ def get_package_extract_template(self):
...
+        if not self.middletier_namespace:
+            raise ConfigurationError(
+                "Running package-analyzer requires middletier namespace to be specified"
+            )
+
+        template = template or self.get_package_analyzer_template()
+
+        self.set_template_parameters(
+            template,
+            THOTH_PACKAGE_ANALYZER_PACKAGE_NAME=package_name,
```
bissenbay: I changed environment variables on [this PR](https://github.com/thoth-station/package-analyzer/pull/4/files) to ```package-analyzer``` as well.

PR #450, opened 2019-08-01
New function to count status of specific jobs identifed by label, in a certain namespace

pacospace:
New function to count jobs per status per label.

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -2145,3 +2145,46 @@ class OpenShift:
             build_cpu = self.parse_cpu_spec(build_cpu)
 
         return build_cpu, build_memory
+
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -2145,3 +2145,48 @@ def _get_build_requests(self, template: dict) -> tuple:
...
+            "pending",
+            "retry",
+            "waiting",
+            "started",
+        ]
+
+        # Initialize
+        jobs_status_count = {}
+        for job_status in status:
+            jobs_status_count[job_status] = 0
```
```python
jobs_status_count = dict.fromkeys(status, 0)
```

pacospace: Updated, thanks :beer:

PR #459, opened 2019-08-14
Do not pin to a specific Kubernetes release

fridex:
Related: https://github.com/thoth-station/graph-sync-scheduler/issues/57 https://github.com/thoth-station/cleanup-job/issues/193


fridex:
Tested with cleanup job, looks like it works.

PR #467, opened 2019-08-22
Add thoth-mypy into .zuul.yaml

fridex:


goern:
recheck

fridex:
recheck

fridex:
Will be replaces with pytest-mypy.

PR #523, opened 2019-11-06
Workflow management

CermakM:

### DIFF
--- a/thoth/common/helpers.py
+++ b/thoth/common/helpers.py
@@ -27,11 +27,15 @@ from typing import Any
 from typing import Dict
 from typing import Generator
 from typing import Optional
+from typing import TypeVar
 
 from contextlib import contextmanager
 
+T = TypeVar("T")
+
 SERVICE_TOKEN_FILENAME = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 SERVICE_CERT_FILENAME = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+
 _DATETIME_FORMAT_STRING = "%Y-%m-%dT%H:%M:%S.%f"
 
 
@@ -108,7 +112,7 @@ def get_service_account_token() -> str:
         ) from exc
 
 
-def to_camel_case(obj: Dict[str, Any]) -> Dict[str, Any]:
+def to_camel_case(obj: T) -> T:
     """Convert dictionary keys to camelCase."""
     if isinstance(obj, dict):
         aux = dict()
@@ -116,12 +120,12 @@ def to_camel_case(obj: Dict[str, Any]) -> Dict[str, Any]:
             new_key = re.sub(r"(?<=.{1})_([a-z])", lambda m: f"{m.group(1).upper()}", key)
             aux[new_key] = to_camel_case(value)
 
-        return aux
+        return aux  # type: ignore
 
     return obj
 
 
-def to_snake_case(obj: Dict[str, Any]) -> Dict[str, Any]:
+def to_snake_case(obj: T) -> T:
     """Convert dictionary keys to snake_case."""
     if isinstance(obj, dict):
         aux = dict()
@@ -129,6 +133,6 @@ def to_snake_case(obj: Dict[str, Any]) -> Dict[str, Any]:
             new_key = re.sub(r"(?<=.{1})([A-Z])", lambda m: f"_{m.group(0)}", key).lower()
             aux[new_key] = to_snake_case(value)
 
-        return aux
+        return aux  # type: ignore
 
     return obj
--- a/thoth/common/helpers.py
+++ b/thoth/common/helpers.py
@@ -19,7 +19,12 @@
 
 import datetime
 import os
+import re
+
 from datetime import timezone
+
+from typing import Any
+from typing import Dict
 from typing import Generator
 from typing import Optional
 
@@ -101,3 +106,29 @@ def get_service_account_token() -> str:
             "Unable to get service account token, please check "
             "that service has service account assigned with exposed token"
         ) from exc
+
+
+def to_camel_case(obj: Dict[str, Any]) -> Dict[str, Any]:
+    """Convert dictionary keys to camelCase."""
+    if isinstance(obj, dict):
+        aux = dict()
+        for key, value in obj.items():
+            new_key = re.sub(r"(?<=.{1})_([a-z])", lambda m: f"{m.group(1).upper()}", key)
+            aux[new_key] = to_camel_case(value)
+
+        return aux
+
+    return obj
+
+
+def to_snake_case(obj: Dict[str, Any]) -> Dict[str, Any]:
+    """Convert dictionary keys to snake_case."""
+    if isinstance(obj, dict):
+        aux = dict()
+        for key, value in obj.items():
+            new_key = re.sub(r"(?<=.{1})([A-Z])", lambda m: f"_{m.group(0)}", key).lower()
+            aux[new_key] = to_snake_case(value)
+
+        return aux
+
+    return obj
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -109,6 +109,8 @@ class OpenShift:
             self.ocp_client = DynamicClient(k8s_client)
             self.in_cluster = False
 
+        self.configuration = self.ocp_client.configuration
+
         self.amun_inspection_namespace = amun_inspection_namespace or os.getenv(
             "THOTH_AMUN_INSPECTION_NAMESPACE"
         )
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -77,14 +77,6 @@ class OpenShift:
                 "installed with openshift extras?"
             ) from exc
 
-        try:
-            import argo.workflows as argo
-        except ImportError as exc:
-            raise ImportError(
-                "Unable to import argo-workflows package. Was thoth-common library "
-                "installed with argo extras?"
-            ) from exc
-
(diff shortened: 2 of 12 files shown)


goern:
@CermakM could you pin openshift to 0.10.0 and `pipenv relock`

thoth/common/openshift.py
```diff
@@ -77,6 +77,14 @@ def __init__(
...
             ) from exc
 
+        try:
+            import argo.workflows as argo
+        except ImportError as exc:
+            raise ImportError(
+                "Unable to import argo-workflows package. Was thoth-common library "
+                "installed with argo extras?"
+            ) from exc
+
```
fridex: We currently do not maintain any extra on this package. I removed them - IMHO having argo as a separate extra does not make sense, we can have `argo+openshift` bundled in one extra if that would be desired (might simplify dependency management later).

thoth/common/openshift.py
```diff
@@ -526,6 +540,43 @@ def _transform_cronjob_to_job(item: Dict[str, Any]) -> Dict[str, Any]:
         job_spec["metadata"]["generateName"] = item["metadata"]["name"] + "-"
         return job_spec
 
+    def _submit_workflow(
+        self,
+        wf: argo.models.V1alpha1Workflow,
+        namespace: Optional[str] = None,
```
fridex: Let's have `namespace` required for this low level API for callee.

thoth/common/openshift.py
```diff
@@ -526,6 +540,43 @@ def _transform_cronjob_to_job(item: Dict[str, Any]) -> Dict[str, Any]:
...
+                name=name,
+                value=value
+            )
+            new_parameters.append(param)
+
+        for p in wf.spec.arguments.parameters:
+            if p.name in parameters:
+                continue  # overridden
+            elif not p.value and not p.default:
+                raise Exception("Required parameter {p.name}")
```
fridex: f-string needed. Probably a separate exception type would be appropriate for this.

thoth/common/openshift.py
```diff
@@ -526,6 +540,43 @@ def _transform_cronjob_to_job(item: Dict[str, Any]) -> Dict[str, Any]:
         job_spec["metadata"]["generateName"] = item["metadata"]["name"] + "-"
         return job_spec
 
+    def _submit_workflow(
+        self,
+        wf: argo.models.V1alpha1Workflow,
+        namespace: Optional[str] = None,
+        parameters: Optional[Dict[str, str]] = None,
+    ):
```
```
) -> None:
```

or, in a better case, a workflow id?

thoth/common/openshift.py
CermakM: Ah I see, I somehow inferred that from the OpenShift/Kubernetes import check above. +1, we can add argo+openshift extras

thoth/common/openshift.py
CermakM: True that! Thanks :+1:

thoth/common/openshift.py
CermakM: Correct, I changed that to `workflow.name` in the meantime and we can discuss potentially more sophisticated approach (i.e. hash or generating id similar to the inspections) if needed.

thoth/common/openshift.py
CermakM: Ack

thoth/common/workflows.py
```diff
@@ -0,0 +1,230 @@
...
+        """Compare workflows for equality."""
+        return self.id == other.id
+
+    def __hash__(self) -> Any:
+        """Compute hash of this Workflow."""
+        return self.to_str().__hash__()
+
+    @classmethod
+    def from_file(cls, fp: Union[str, Path], fmt: str = "yaml") -> "Workflow":
+        """Create a Workflow from a file."""
```
fridex: It would be better to follow the current approach - store argo workflows in OpenShift as templates not to bundle templates with source code.

CermakM: OK, I will do it that way in the actual implementation and I'll leave the function just for convenience (quite useful for testing stuff IMHO), thanks for the advice!

goern: should every workflow instance be created from an WorkflowTemplate?

CermakM: I don't think so. We've briefly touched the subject with @fridex and we will most likely have the top level workflow wrapped in an OpenShift Template.

goern:
recheck

thoth/common/helpers.py
```diff
@@ -101,3 +106,29 @@ def get_service_account_token() -> str:
             "Unable to get service account token, please check "
             "that service has service account assigned with exposed token"
         ) from exc
+
+
+def to_camel_case(obj: Dict[str, Any]) -> Dict[str, Any]:
+    """Convert dictionary keys to camelCase."""
+    if isinstance(obj, dict):
```
fridex: The function signature already states `obj` is a dictionary. It would be better to move this check down bellow before the actual recursion is done.

thoth/common/helpers.py
```diff
@@ -101,3 +106,29 @@ def get_service_account_token() -> str:
...
+        aux = dict()
+        for key, value in obj.items():
+            new_key = re.sub(r"(?<=.{1})_([a-z])", lambda m: f"{m.group(1).upper()}", key)
```
fridex: The f-string is not required.

thoth/common/helpers.py
```diff
@@ -101,3 +106,29 @@ def get_service_account_token() -> str:
...
+            aux[new_key] = to_camel_case(value)
+
+        return aux
+
+    return obj
+
+
+def to_snake_case(obj: Dict[str, Any]) -> Dict[str, Any]:
+    """Convert dictionary keys to snake_case."""
+    if isinstance(obj, dict):
```
fridex: The same as for `to_camel_case`.

thoth/common/workflows.py
```diff
@@ -0,0 +1,234 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Workflow management for Thoth."""
+
+import logging
+import json
+import random
+import re
```
fridex: Haven't check all imports, some of them look unused - like this `re` import.

thoth/common/workflows.py
```diff
@@ -0,0 +1,234 @@
...
+    def name(self) -> Union[str, None]:
+        """Get Workflow name."""
+        name: Union[str, None] = getattr(self.metadata, "name", None)
+        return name
+
+    @property
+    def id(self) -> str:
+        """Get Workflow ID."""
+        digest: str = abs(self.__hash__())
+        return f"workflow{'-' + self.name if self.name else ''}-{digest}"
```
fridex: It might be a good idea to reuse digest creation from the openshift instance to unify digest creation (you can make it public):

https://github.com/thoth-station/common/blob/c21cd41a3a5f52c756225f34b7e7f4cc16f21bfb/thoth/common/openshift.py#L1137-L1139

thoth/common/workflows.py
```diff
@@ -0,0 +1,234 @@
...
+    @classmethod
+    def from_file(cls, fp: Union[str, Path], fmt: str = "yaml") -> "Workflow":
+        """Create a Workflow from a file."""
+        fmt = fmt.lower()
+
+        wf_path = Path(fp)
+
+        wf: Dict[str, Any]
+        if fmt == "yaml":
+            wf = yaml.safe_load(wf_path.read_text())
```
fridex: JSON is a subset of YAML, you can discard `fmt` argument and use just `yaml.safe_load`:

```python
>>> import yaml
>>> yaml.safe_load('{"foo": 1, "bar": ["baz"]}')
{'foo': 1, 'bar': ['baz']}
```

thoth/common/workflows_test.py
```diff
@@ -0,0 +1,46 @@
+"""Workflows test suite."""
```
fridex: It would be a good idea to move the test suite out of the prod code. You can place them under `tests/` directory as in the case of other libraries in Thoth. Bots will run the test suite on each PR. I created the `tests/` dir in https://github.com/thoth-station/common/pull/529

thoth/common/workflows.py
CermakM: Ah, correct, leftover from moving the helper function to helpers module :) Thanks!

thoth/common/workflows.py
CermakM: That's neat!

thoth/common/helpers.py
CermakM: I see your point. I do have to check for the dict at the beginning tho IMHO. However, I can refactor the function to contain an inner recursive function with a proper signature to make it a bit clearer.

fridex: You can just change the signature to:

```python
def to_camel_case(obj: Any) -> Any:
```

fridex: You can just change the signature to:

```python
def to_camel_case(obj: Any) -> Any:
```

CermakM: Aight, will do.. maybe even better to use a generic to emphasize that the return type is the same as the input `obj` type?

```python
def to_camel_case(obj: T) -> T:
```

fridex: Sounds good :+1:

thoth/common/workflows_test.py
CermakM: Done :) Thanks!

CermakM:
Thanks for the thorough review @fridex !

PR #530, opened 2019-11-13
Do not propagate private property on to_dict()

fridex:


fridex:
This addresses pytest issues in https://github.com/thoth-station/adviser/pull/504 and https://github.com/thoth-station/adviser/pull/503

thoth/common/config/runtime_environment.py
```diff
@@ -95,6 +95,8 @@ def from_dict(cls, dict_: Optional[Dict[Any, Any]] = None) -> "RuntimeEnvironmen
     def to_dict(self, without_none: bool = False) -> Dict[str, Any]:
         """Convert runtime environment configuration to a dict representation."""
         dict_ = attr.asdict(self)
+        # Do not propagate private property.
+        dict_.pop("_fully_specified", None)
```
goern: why not del that key from the dict?

PR #535, opened 2019-11-15
Fix error due to ambiguous template resources

CermakM:
This error occurs with `openshift~=10.0` due to the `Template` resource being present under `name=templates` and `name=processedtemplates`.


goern:
LGTM

PR #539, opened 2019-11-19
Enable Sentry integrations

fridex:


fridex:
Discarding WIP flag, tested in test environment with user-api, management-api and metrics-exporter.

PR #596, opened 2020-01-06
Be more aggressive with busy wait

fridex:


fridex:
I'm wondering if this is the right fix - the issue with non-existing adviser workload happening from time to time on user-api might be caused by workload-operator. User-api asks for job, it does not find it, so it asks for cm which is not found as workload operator meantime transformed cm into job. Let's see if this adresses the issue...

fridex:
Anyway - this is a quick&dirty fix until we start using argo.

PR #603, opened 2020-01-08
Migration to Argo workflow for Adviser

pacospace:

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -144,11 +144,19 @@ class OpenShift:
                 "TLS verification when communicating with k8s/okd master is disabled"
             )
 
-        self.use_argo =  use_argo or bool(int(os.getenv("THOTH_USE_ARGO", 0)))
+        self.use_argo = use_argo or bool(int(os.getenv("THOTH_USE_ARGO", 0)))
+
+        self.workflow_manager = None
 
         if self.use_argo:
             _LOGGER.info(
                 "Using Argo Workflow to run jobs"
+            )
+            import workflows
+            self.workflow_manager = workflows.WorkflowManager(
+                ocp_config={
+                    "kubernetes_verify_tls": self.kubernetes_verify_tls
+                    }
             )
 
     @property
@@ -1509,7 +1517,6 @@ class OpenShift:
                 labels={"component": "adviser"},
             )
 
-        import workflows
         adviser_id = job_id or self.generate_id("adviser")
         template_parameters = {}
         template_parameters["THOTH_ADVISER_JOB_ID"] = adviser_id
@@ -1518,22 +1525,21 @@ class OpenShift:
         template_parameters["THOTH_ADVISER_LIBRARY_USAGE"] = json.dumps(library_usage)
         template_parameters["THOTH_ADVISER_REQUIREMENTS_FORMAT"] = "pipenv"
         template_parameters["THOTH_ADVISER_RECOMMENDATION_TYPE"] = recommendation_type
-        template_parameters["THOTH_ADVISER_RUNTIME_ENVIRONMENT"] = runtime_environment
-        template_parameters["THOTH_ADVISER_LIMIT"] = limit
-        template_parameters["THOTH_ADVISER_COUNT"] = count
-        template_parameters["THOTH_ADVISER_LIMIT_LATEST_VERSIONS"] = limit_latest_versions
+        template_parameters["THOTH_ADVISER_RUNTIME_ENVIRONMENT"] = json.dumps(runtime_environment)
+
+        if limit is not None:
+            template_parameters["THOTH_ADVISER_LIMIT"] = limit
+
+        if count is not None:
+            template_parameters["THOTH_ADVISER_COUNT"] = count
+
+        if limit_latest_versions is not None:
+            template_parameters["THOTH_ADVISER_LIMIT_LATEST_VERSIONS"] = limit_latest_versions
 
         workflow_parameters = {}
-        workflow_parameters["THOTH_DOCUMENT_ID"] = f"advise-{adviser_id}"
-
-        workflow_manager = workflows.WorkflowManager(
-            ocp_config={
-                "kubernetes_verify_tls": self.kubernetes_verify_tls
-                }
-        )
 
         return self._schedule_workflow(
-            workflow=workflow_manager.submit_adviser_workflow,
+            workflow=self.workflow_manager.submit_adviser_workflow,
             parameters={
                 "adviser_id": adviser_id,
                 "template_parameters": template_parameters,
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -41,7 +41,6 @@ from .helpers import (
     _get_incluster_token_file,
     _get_incluster_ca_file,
 )
-from .workflows import WorkflowManager
 
 urllib3.disable_warnings()
 _LOGGER = logging.getLogger(__name__)
@@ -64,6 +63,7 @@ class OpenShift:
         kubernetes_api_url: Optional[str] = None,
         kubernetes_verify_tls: bool = True,
         openshift_api_url: Optional[str] = None,
+        use_argo: bool = False,
         token: Optional[str] = None,
         token_file: Optional[str] = None,
         cert_file: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


thoth/common/openshift.py
```diff
@@ -1479,19 +1488,37 @@ def schedule_adviser(
...
+        workflow_parameters = {}
+        workflow_parameters["THOTH_DOCUMENT_ID"] = f"advise-{adviser_id}"
+
+        workflow_manager = WorkflowManager(
+            ocp_config={
+                "kubernetes_verify_tls": self.kubernetes_verify_tls
+                }
         )
 
+        return self._schedule_workflow(
```
fridex: Can we introduce a new environment variable `THOTH_USE_ARGO`, which would (if set to `1`) cause triggering the new implementation in thoth-common that uses argo? If set to `0` (or not set), the old implementation that uses our workload-operator+graph-sync-scheduler would be used. This way we can transparently switch and/or fallback to the old implementation until things stabilize.

thoth/common/openshift.py
```diff
@@ -143,6 +144,15 @@ def __init__(
                 "TLS verification when communicating with k8s/okd master is disabled"
             )
 
+        self.use_argo = bool(
+            int(os.getenv("THOTH_USE_ARGO", 0)) and use_argo
```
fridex: This means both will need to be set - environment variable and the parameter passed in the constructor in order to run argo workflows. It would be better to control this purely on the environment variable or let the parameter override the value:

```python
    self.use_argo = use_argo or int(os.getenv("THOTH_USE_ARGO", 0))

Comment (pacospace) [reply thoth/common/openshift.py]: Updated, thanks :)

File: thoth/common/openshift.py [hunk 3]
```diff
@@ -1480,18 +1496,51 @@ def schedule_adviser(
...
+        template_parameters["THOTH_ADVISER_LIBRARY_USAGE"] = json.dumps(library_usage)
+        template_parameters["THOTH_ADVISER_REQUIREMENTS_FORMAT"] = "pipenv"
+        template_parameters["THOTH_ADVISER_RECOMMENDATION_TYPE"] = recommendation_type
+        template_parameters["THOTH_ADVISER_RUNTIME_ENVIRONMENT"] = runtime_environment
+        template_parameters["THOTH_ADVISER_LIMIT"] = limit
+        template_parameters["THOTH_ADVISER_COUNT"] = count
+        template_parameters["THOTH_ADVISER_LIMIT_LATEST_VERSIONS"] = limit_latest_versions
+
+        workflow_parameters = {}
+        workflow_parameters["THOTH_DOCUMENT_ID"] = f"advise-{adviser_id}"
```
fridex: Just:

```python
        workflow_parameters["THOTH_DOCUMENT_ID"] = adviser_id
```

Or will it be something like `advise-adviser-401a385e`?

thoth/common/openshift.py
```diff
@@ -1480,18 +1496,51 @@ def schedule_adviser(
...
+
+        workflow_manager = workflows.WorkflowManager(
```
fridex: I think this will be used more times, maybe worth to make it property or instance variable?

thoth/common/openshift.py
```diff
@@ -1480,18 +1496,51 @@ def schedule_adviser(
...
+            ocp_config={
+                "kubernetes_verify_tls": self.kubernetes_verify_tls
+                }
         )
 
+        return self._schedule_workflow(
```
fridex: Just to make sure, this will result in `adviser_id` being returned?

pacospace: Yes it will return the `adiviser_id` or any `workflow_id` depending on the component selected (e.g. adviser, inspection, ...)

thoth/common/openshift.py
```diff
@@ -1480,18 +1496,51 @@ def schedule_adviser(
...
+        template_parameters["THOTH_ADVISER_JOB_ID"] = adviser_id
+        template_parameters["THOTH_ADVISER_REQUIREMENTS"] = application_stack["requirements"]
+        template_parameters["THOTH_ADVISER_REQUIREMENTS_LOCKED"] = application_stack["requirements_lock"]
+        template_parameters["THOTH_ADVISER_LIBRARY_USAGE"] = json.dumps(library_usage)
+        template_parameters["THOTH_ADVISER_REQUIREMENTS_FORMAT"] = "pipenv"
+        template_parameters["THOTH_ADVISER_RECOMMENDATION_TYPE"] = recommendation_type
+        template_parameters["THOTH_ADVISER_RUNTIME_ENVIRONMENT"] = runtime_environment
+        template_parameters["THOTH_ADVISER_LIMIT"] = limit
+        template_parameters["THOTH_ADVISER_COUNT"] = count
+        template_parameters["THOTH_ADVISER_LIMIT_LATEST_VERSIONS"] = limit_latest_versions
```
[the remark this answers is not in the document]
pacospace: Can we put the default values from the template?
```
template_parameters["THOTH_ADVISER_RUNTIME_ENVIRONMENT"] = runtime_environment
        template_parameters["THOTH_ADVISER_LIMIT"] = limit or 1000
        template_parameters["THOTH_ADVISER_COUNT"] = count or 1
        template_parameters["THOTH_ADVISER_LIMIT_LATEST_VERSIONS"] = limit_latest_versions or 5
```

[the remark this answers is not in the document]
fridex: IMHO, it would be better to keep them at one place - to use the default values from template, you can guard assignment with if statements:

```python
if limit is not None:
    template_parameters["THOTH_ADVISER_LIMIT"] = limit
```

`template_parameters` will not hold `THOTH_ADVISER_LIMIT` and thus the default values from the template will be used. WDYT?

thoth/common/openshift.py
pacospace: From the workflow in adviser workflow template, we have objects:
  - apiVersion: argoproj.io/v1alpha1
    kind: Workflow
    metadata:
      name: "advise-${THOTH_ADVISER_JOB_ID}"
      annotations:
        thoth-station.ninja/template-version: 0.5.0
      labels:
        app: thoth
        component: adviser
    spec:
      serviceAccountName: argo
      podGC:
        strategy: OnWorkflowSuccess
      volumes:
        - name: cache
          emptyDir: {}
      entrypoint: advise
      arguments:
        parameters:
          - name: THOTH_ADVISER_JOB_ID
            value: "${THOTH_ADVISER_JOB_ID}"
          - name: THOTH_DOCUMENT_ID
            value: "advise-${THOTH_ADVISER_JOB_ID}"

fridex: To integrate this correctly with graph-sync, we will probably need to discard `advise-` prefix - why is it needed?

The point here I'm making - graph-sync-job gets prefix of the document name to check what syncing handler should be run:

https://github.com/thoth-station/storages/blob/master/thoth/storages/sync.py#L457-L465

If the resulting document will be named `advise-adviser-401a385e` (that is IMHO not nice name), graph-sync-job fails to find the correct handler.

pacospace: I agree, so I will adjust the adviser workflow template as well :) Thanks!!

thoth/common/openshift.py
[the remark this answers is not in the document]
pacospace: Actually a much better solution, Thanks :beer:

thoth/common/openshift.py
```diff
@@ -143,6 +144,19 @@ def __init__(
...
             )
 
+        self.use_argo = use_argo or bool(int(os.getenv("THOTH_USE_ARGO", 0)))
+
+        if self.use_argo:
+            _LOGGER.info(
+                "Using Argo Workflow to run jobs"
+            )
+            import workflows
+            self.workflow_manager = workflows.WorkflowManager(
```
fridex: Please add:
```
self.workflow_manager = None
```

before if so the property exists, but is not set.

thoth/common/openshift.py
```diff
@@ -1480,17 +1502,48 @@ def schedule_adviser(
...
+
+        adviser_id = job_id or self.generate_id("adviser")
+        template_parameters = {}
+        template_parameters["THOTH_ADVISER_JOB_ID"] = adviser_id
+        template_parameters["THOTH_ADVISER_REQUIREMENTS"] = application_stack["requirements"]
+        template_parameters["THOTH_ADVISER_REQUIREMENTS_LOCKED"] = application_stack["requirements_lock"]
+        template_parameters["THOTH_ADVISER_LIBRARY_USAGE"] = json.dumps(library_usage)
+        template_parameters["THOTH_ADVISER_REQUIREMENTS_FORMAT"] = "pipenv"
+        template_parameters["THOTH_ADVISER_RECOMMENDATION_TYPE"] = recommendation_type
+        template_parameters["THOTH_ADVISER_RUNTIME_ENVIRONMENT"] = runtime_environment
```
fridex: I think runtime environment needs to be serialized as well:

```
        template_parameters["THOTH_ADVISER_RUNTIME_ENVIRONMENT"] = json.dumps(runtime_environment)
```

PR #620, opened 2020-01-16
Try to avoid timing issues between job and configmap creation

fridex:


thoth/common/openshift.py
```diff
@@ -479,10 +479,9 @@ def get_job_status_report(self, job_id: str, namespace: str) -> Dict[str, Option
                     "state": "registered",
                 }
             except Exception:
-                pass
-
-            # Raise the original exception as the given pod was not found based on job id.
-            raise
+                # Try to obtain the job once again - this can avoid timing issues when job
+                # is just created and we are asking for the job id.
+                job_id = self._get_pod_id_from_job(job_id, namespace)
```
harshad16: Isnt this tricky :) , what if the populating takes more time to start a job(maybe in creating container) . this will fail again

fridex: I'm not sure why we encounter the issue. My understanding based on what's happening on workload-operator:

1. take a CM
2. turn CM into JOB
3. delete CM

On user-api we do:

1. get job
2. if not found, get CM
3. if not found, error

If we hit workload-operator between 1. and 2., we end up with the exception raised (3. step on user-api). My assumption is that if we try to get job status again, we could retrieve "pending" state of the job.

As said, I'm not that sure - it might be also the time needed to sync on Kubernetes API. I think it's worth to try.

harshad16: thanks for the explanation, lets try it out :+1:

fridex:
@harshad16 I did some testing, it looks like this patch addressed the issue :tada: :+1:

PR #628, opened 2020-01-17
Rename template used by workload opeartor

pacospace:
Related-To: https://github.com/thoth-station/adviser/pull/769

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -918,7 +918,7 @@ class OpenShift:
                 "Infra namespace is required to gather solver template when running solver"
             )
 
-        template = self._get_template("template=solver")
+        template = self._get_template("template=solver-workload-operator")
 
         # Get only one solver - the solver that was requested.
         solver_entry = None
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -1633,7 +1633,7 @@ def get_adviser_template(self) -> Dict[str, Any]:
                 "Infra namespace is required to gather adviser template when running it"
             )
 
-        return self._get_template("template=adviser")
+        return self._get_template("template=adviser-argo")
```
fridex: `if self.use_argo`?

pacospace: right, `adviser-workload-operator`

pacospace: We can switch the other way around, as Argo will be the default, as suggested by @goern. WDYT?

fridex: I think we missed each other:

```python
if self.use_argo:
        return self._get_template("template=adviser")
else:
        return self._get_template("template=adviser-workload-operator")
```

To stay compatible with the env variable switch.

pacospace: Sorry, I didn't get that indeed :) Let's discuss it on Monday :)

pacospace: Updated, as discussed. Thanks! :beers:

pacospace:
recheck

PR #633, opened 2020-01-21
Sentryfilter

saisankargochhayat:
I have tested this with setting the THOTH_SENTRY_IGNORE_EXCEPTION=connexion.decorators.validation,ValueError
Both for hint types which have exc_info data and types which have just log_record data.


thoth/common/logging.py
```diff
@@ -148,6 +148,19 @@ def _get_sentry_integrations() -> List[object]:
...
 
+def before_send_handler(event, hint):
+    """ Ignores the exceptions passed in as as the environment variable in a comma separated manner.
+    """
+    ignored_exceptions = os.environ("THOTH_SENTRY_IGNORE_EXCEPTION")
+    print(ignored_exceptions)
+    if ignored_exceptions:
+        if 'exc_info' in hint:
+            exc_type, exc_value, tb = hint['exc_info']
+        for exception in ignored_exceptions.split(','):
```
fridex: This block should be executed only for if exc_info is present - we can send also errors into sentry using `_LOGGER.error()` which will not hold any exception.

thoth/common/logging.py
```diff
@@ -148,6 +148,19 @@ def _get_sentry_integrations() -> List[object]:
...
+            if isinstance(exc_value, exception):
```
fridex: `exception` will be of type string (not type, but the value of type). What is the type of exc_value?

thoth/common/logging.py
```diff
@@ -148,6 +148,19 @@ def _get_sentry_integrations() -> List[object]:
 
     return integrations
 
+def before_send_handler(event, hint):
+    """ Ignores the exceptions passed in as as the environment variable in a comma separated manner.
+    """
+    ignored_exceptions = os.environ("THOTH_SENTRY_IGNORE_EXCEPTION")
```
fridex: `os.environ` is a dict, access its element using `os.environ["THOTH_SENTRY_IGNORE_EXCEPTION"]`, or better for this case, use `os.getenv("THOTH_SENTRY_IGNORE_EXCEPTION")` which will return `None` or the environment variable value.

thoth/common/logging.py
saisankargochhayat: Thanks! :smile:

thoth/common/logging.py
saisankargochhayat: `{'log_record': <LogRecord: connexion.decorators.validation, 40, /home/sai/.local/share/virtualenvs/user-api-ZDh_dRcv/lib/python3.6/site-packages/connexion/decorators/validation.py, 204, "http://localhost:5000/api/v1/advise/python?recommendation_type=stable validation error: 3 is not of type 'string' - 'runtime_environment.python_version'">}`
This is what the hint value looks like, its of the type - 'dict'

fridex: I was referring to variable `exception` which you use in the loop to iterate over ignored exceptions.

What about the `event` passed? What information does it carry?

thoth/common/logging.py
```diff
@@ -148,6 +148,26 @@ def _get_sentry_integrations() -> List[object]:
...
+            exc_type, exc_value, tb = hint['exc_info']
+            for exception in exceptions_split:
+                if exception == exc_type.__name__:
+                    _LOGGER.info(hint)
+                    return None
+        elif 'log_record' in hint:
+            log_record = hint['log_record'].__dict__
+            for exception in exceptions_split:
+                if exception == log_record['name']:
+                    _LOGGER.info(hint)
```
saisankargochhayat: This for now doesn't send to sentry.

thoth/common/logging.py
saisankargochhayat: In the current method, `exception` is the variable that iterates over the exceptions to be ignored. And the event has various details used by sentry, like stack trace, packages installed, error name. (link to a sample event - https://paste.ofcode.org/3WMTFq9zMLzTyku29EMutR
api called event - https://pastecode.xyz/view/32f334d2)

README.rst
```diff
@@ -63,3 +63,11 @@ to Sentry:
...
 application itself.
 
+
+If you want some certain classes of error to be not reported to Sentry. You can
+can provide ``THOTH_SENTRY_IGNORE_EXCEPTION`` environment variable which holds
+the comma separated list of exception classes to be ignored when reporting to Sentry:
+
+.. code-block:: console
+
+  THOTH_SENTRY_IGNORE_EXCEPTION="connexion.decorators.validation,ValueError"
```
fridex: Is the comma here intentional? Is it delimiting module and exception class or is it just a typo?

saisankargochhayat: It's there to facilitate a split between different exception classes.

fridex: So one exception is `connexion.decorators.validation` and another one is `ValueError`?

thoth/common/logging.py
```diff
@@ -149,6 +149,29 @@ def _get_sentry_integrations() -> List[object]:
     return integrations
 
 
+def before_send_handler(event, hint):
+    """Filter the errors caught before sending to Sentry.
+
+    This function ignores the exceptions passed in as a environment variable in a comma separated manner.
+    """
+    ignored_exceptions = os.getenv('THOTH_SENTRY_IGNORE_EXCEPTION')
+    if ignored_exceptions:
```
fridex: To reduce code complexity:

```python
    ignored_exceptions = os.getenv('THOTH_SENTRY_IGNORE_EXCEPTION')
    if not ignored_exceptions:
        return event

    # Complementary code here...
````

File: thoth/common/logging.py [hunk 7]
```diff
@@ -149,6 +149,29 @@ def _get_sentry_integrations() -> List[object]:
...
+        exceptions_split = ignored_exceptions.split(',')
+        if 'exc_info' in hint:
+            exc_type, exc_value, tb = hint['exc_info']
+            for exception in exceptions_split:
+                if exception == exc_type.__name__:
+                    _LOGGER.info(hint)
```
fridex: No `_LOGGER.info` should be needed here - the exception will be printed to standard error as expected but it will not be sent to sentry.

thoth/common/logging.py
```diff
@@ -149,6 +149,29 @@ def _get_sentry_integrations() -> List[object]:
...
+
+    This function ignores the exceptions passed in as a environment variable in a comma separated manner.
+    """
+    ignored_exceptions = os.getenv('THOTH_SENTRY_IGNORE_EXCEPTION')
+    if ignored_exceptions:
+        exceptions_split = ignored_exceptions.split(',')
+        if 'exc_info' in hint:
+            exc_type, exc_value, tb = hint['exc_info']
+            for exception in exceptions_split:
+                if exception == exc_type.__name__:
```
fridex: Will this work if two libraries will implement an exception with a same name?

Let's say we will use two libraries `lib1` and `lib2`. Both implement an exception called `MyException` in `lib1.exceptions.MyException` and `lib2.exceptions.MyException`. Can we distinguish these two exceptions?

thoth/common/logging.py
saisankargochhayat: Will update the function.

README.rst
saisankargochhayat: Yes

thoth/common/logging.py
saisankargochhayat: Okay will remove it.

thoth/common/logging.py
saisankargochhayat: I think we would need to add `lib1.exceptions.MyException,lib2.exceptions.MyException`, both to the env variables to ignore them. Alternatively we could check if MyException is in the string `lib1.exceptions.MyException`, but that might ignore more errors than we want to I guess.

saisankargochhayat: Is that what you were asking?

fridex: > We would need to add `lib1.exceptions.MyException,lib2.exceptions.MyException`, both to the env variables to ignore them.

What if we want to ignore just `MyException` from `lib1`, but keep reporting `MyException` from `lib2`. Will it be possible?

saisankargochhayat: Yes, it should be, as here for example, we are checking for 'connexion.decorators.validation', this doesn't ignore 'xyz.validation'. 
However for generic python errors like ValueError, we are checking for exc_info type, not value, that is here in this example class type - 
`{'exc_info': (<class 'ValueError'>, ValueError('My really new error',), <traceback object at 0x7f2938edd0c8>), 'log_record': <LogRecord: test, 40, app.py, 10, "exception raised">}`
Should we catch the value? Btw that would only prevent handled errors, and the specific messages need to be added to env variables, which doesn't seem like a good idea.

fridex: No exception value needed - we need to capture the correct exception type from the right module.

Could you verify this works as expected with your patch?

```console
$ cat lib1.py                                                                
class Test(Exception):
    pass
                                                                                
$ cat lib2.py                                                                
class Test(Exception):
    pass
                                                                                
$ cat test.py                                                                
import sys

from lib1 import Test as Test1
from lib2 import Test as Test2

try:
    raise Test1
except Exception:
    print(sys.exc_info()[0].__name__)
                                                                                
$ python3 test.py                                                            
Test
```

saisankargochhayat: So, the str(exc_type) returns - `<class 'lib1.Test'> `
Where as `exc_type.__name__`  returns 'Test' i.e. doesn't carry which module it is from. 
Should we do str(exc_type) and then some sort of string formatting?

saisankargochhayat:
```
print(sys.exc_info())

print(str(sys.exc_info()[0]))

print(sys.exc_info()[0].__name__)

print(sys.exc_info()[0].__qualname__)
```

Gives - 
````
(<class 'lib1.Test'>, Test(), <traceback object at 0x7f3fc9fa0ec8>)

<class 'lib1.Test'>

Test

Test

Comment (fridex) [reply thoth/common/logging.py]: No string passing/ formatting needed. Can you come up with some other solution? 

Hint: `getattr` and `str.rsplit`

Comment (fridex) [reply thoth/common/logging.py]: The two methods above are your friends 🙂

Comment (fridex) [reply thoth/common/logging.py]: Or simply `inspect.getmodule`.

Comment (saisankargochhayat) [reply thoth/common/logging.py]: Something like this?
`print(getattr(exc_type, "__module__")+'.'+exc_type.__name__)`
`lib1.Test`
These are all the attributes exc_type class has - 
`{'__module__': 'lib1', '__weakref__': <attribute '__weakref__' of 'Test' objects>, '__doc__': None}
`
Doesn't str.rsplit fall under string formatting? :stuck_out_tongue:

Comment (saisankargochhayat) [reply thoth/common/logging.py]: Does this make sense?
`print(getattr(exc_type, "__module__")+'.'+exc_type.__name__)`
inspect.getmodule gives (plus we need to import inspect) - 
`<module 'lib1' from '/home/sai/Downloads/thoth-station/common/lib1.py'>
`
These are all the attr of exc_type - 
`{'__module__': 'lib1', '__weakref__': <attribute '__weakref__' of 'Test' objects>, '__doc__': None}
`

Comment (saisankargochhayat) [reply thoth/common/logging.py]: But with this approach for errors like ValueError, we need to specify `builtins.ValueError` in the env variable.

Comment (fridex) [reply thoth/common/logging.py]: > Something like this?
> `print(getattr(exc_type, "__module__")+'.'+exc_type.__name__)`
> `lib1.Test`
> These are all the attributes exc_type class has -
> `{'__module__': 'lib1', '__weakref__': <attribute '__weakref__' of 'Test' objects>, '__doc__': None} `

Using the `__module__` attribute directly with `__name__` sounds good to me. :+1: 

> Doesn't str.rsplit fall under string formatting?

I don't this so: https://docs.python.org/3/library/string.html

File: thoth/common/logging.py [hunk 9]
```diff
@@ -149,6 +149,28 @@ def _get_sentry_integrations() -> List[object]:
...
+    This function ignores the exceptions passed in as a environment variable in a comma separated manner.
+    """
+    ignored_exceptions = os.getenv('THOTH_SENTRY_IGNORE_EXCEPTION')
+    if not ignored_exceptions:
+        return event
+    exceptions_split = ignored_exceptions.split(',')
+    if 'exc_info' in hint:
+        exc_type, exc_value, tb = hint['exc_info']
+        for exception in exceptions_split:
+            if exception.strip() == exc_type.__name__:
```
fridex: This operation will be done very often, so no need to do any additional operations like strip - the docs are explicitly saying how the string should be formatted.

fridex: This logic can be done in `init_logging` and it can set up global variables so we unroll these operations out of the actual `before_send_handler` to speed up these error messages handling and fail on invalid configuration before the application starts:

```python
parts = exception.rsplit(maxsplit=1)
if len(parts) != 2:
    # error about wrong configuration supplied
exc_module, exc_name = parts

```

The core checking logic in `before_send` can then iterate over global variables as initialized in `init_logging` and perform something like this:

```python

if exc_type.__name__ == exc_name and exc_type.__module__ == exc_module:
    return None
```

WDYT?

saisankargochhayat: Okay will remove this.

saisankargochhayat: So sometimes hint doesn't have exc_info, in that case we fall back to log_record. I think it would be a good idea to have that logic in the before_send_handler, also makes it a little more modular.

README.rst
fridex: please adjust `ValueError` to `builtins.ValueError` in that case.

thoth/common/logging.py
saisankargochhayat: @fridex I have moved it to the init_logging method and modified the before_send_handler accordingly. Could you have a look?

thoth/common/logging.py
```diff
@@ -214,6 +237,17 @@ def init_logging(
         for logger in ignored_loggers.split(","):
             ignore_logger(logger)
 
+    ignored_exceptions = os.getenv("THOTH_SENTRY_IGNORE_EXCEPTION")
+    if ignored_exceptions:
+        exceptions_split = ignored_exceptions.split(',')
+        for exception in exceptions_split:
+            exception_parts = exception.rsplit('.', maxsplit=1)
+            if len(exception_parts) != 2:
+                root_logger.exception(
```
fridex: Log error here, you don't have any exception context here (no exception being handled). Also, the assignment bellow can fail as you don't have 2 parts:

```
            exc_module, exc_name = exception_parts

```

You can simply continue with the loop.

thoth/common/logging.py
```diff
@@ -214,6 +237,17 @@ def init_logging(
...
+                    "Configuration for exceptions to be ignored not set correctly.")
+            exc_module, exc_name = exception_parts
+            _IGNORED_EXCEPTIONS.append([exc_module, exc_name])
```
fridex: To optimize a bit, you can append tupple:

```python
            _IGNORED_EXCEPTIONS.append((exc_module, exc_name))
```

thoth/common/logging.py
```diff
@@ -149,6 +150,28 @@ def _get_sentry_integrations() -> List[object]:
...
+    if len(_IGNORED_EXCEPTIONS) == 0:
+        return event
+    if 'exc_info' in hint:
+        exc_type, exc_value, tb = hint['exc_info']
+        for exception in _IGNORED_EXCEPTIONS:
+            module, name = exception
+            if module == getattr(exc_type, "__module__") and name == exc_type.__name__:
+                return None
+    elif 'log_record' in hint:
+        log_record = hint['log_record'].__dict__
```
fridex: Why `__dict__` here? You can directly access the property:

```python
hint['log_record'].name
```

README.rst
```diff
@@ -63,3 +63,14 @@ to Sentry:
 This is helpful if you want to report errors to users but not to Thoth
 application itself.
 
+
+If you want some certain classes of error to be not reported to Sentry. You can
```
fridex: Probably:

```not to be reported```

(but my english is not pro :) )

File: thoth/common/logging.py (hunk 12 above)
Comment (saisankargochhayat) [reply thoth/common/logging.py]: Makes sense.

File: thoth/common/logging.py (hunk 10 above)
Comment (saisankargochhayat) [reply thoth/common/logging.py]: I have put it now under the if check, so it shouldn't give an error now.

File: thoth/common/logging.py [hunk 14]
```diff
@@ -214,6 +237,18 @@ def init_logging(
...
+    if ignored_exceptions:
+        exceptions_split = ignored_exceptions.split(',')
+        for exception in exceptions_split:
+            exception_parts = exception.rsplit('.', maxsplit=1)
+            if len(exception_parts) == 2:
+                exc_module, exc_name = exception_parts
+                _IGNORED_EXCEPTIONS.append((exc_module, exc_name))
+            else:
+                root_logger.error(
+                    "Configuration for exceptions to be ignored not set correctly.")
```
fridex: Generally, it's good to report what's wrong/what code could not parse. It helps with debugging later on. :)

saisankargochhayat: Let me add that here, one min!

fridex: You are challenged with bots :robot: They merge on their own once the PR is accepted. :)

PR #636, opened 2020-01-24
Thamos workflow for GitHub App Qeb-Hwt

pacospace:

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -148,10 +148,9 @@ class OpenShift:
 
         self.workflow_manager = None
         if self.use_argo:
-            _LOGGER.info(
-                "Using Argo Workflow to run jobs"
-            )
+            _LOGGER.info("Using Argo Workflow to run jobs")
             from .workflows import WorkflowManager
+
             self.workflow_manager = WorkflowManager(openshift=self)
 
     @property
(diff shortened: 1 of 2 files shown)


thoth/common/openshift.py
```diff
@@ -1933,8 +2003,40 @@ def schedule_kebechet_run_url(
             labels={"component": "kebechet"},
         )
 
+    def schedule_thamos_workflow(
+        self, check_run_id: int, repo_url: str, commit_sha: str, installation_id: int
+    ) -> str:
+        """Schedule Thamos Advise Workflow for Qeb-Hwt GitHub App.."""
+        if not self.use_argo:
+            return NotImplementedError
```
fridex: I think this should be:

```
            raise NotImplementedError
```

thoth/common/openshift.py
```diff
@@ -1933,8 +2003,40 @@ def schedule_kebechet_run_url(
...
+
+        workflow_id = self.generate_id("")
```
fridex: It might be a good idea to accept `None` in the `generate_id` method to avoid concatenation with empty string. How will workflow be named?

pacospace: ```"thamos-advise-check-run-${EVENT_ID}"```

fridex: Sounds good to me, just that default value to keep sources clean :) #nitpicking

PR #647, opened 2020-01-27
Added log error

saisankargochhayat:
Closes https://github.com/thoth-station/user-api/issues/740
Added missing commit.

### DIFF
--- a/thoth/common/logging.py
+++ b/thoth/common/logging.py
@@ -247,7 +247,7 @@ def init_logging(
                 _IGNORED_EXCEPTIONS.append((exc_module, exc_name))
             else:
                 root_logger.error(
-                    "The following exception couldn't be parsed - "+exception)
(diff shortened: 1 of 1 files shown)


thoth/common/logging.py
```diff
@@ -247,7 +247,7 @@ def init_logging(
                 _IGNORED_EXCEPTIONS.append((exc_module, exc_name))
             else:
                 root_logger.error(
-                    "Configuration for exceptions to be ignored not set correctly.")
+                    "The following exception couldn't be parsed - "+exception)
```
fridex: Use lazy evaluation of the string:

```python
"The following configuration for ignoring exception couldn't be parsed: %r ", exception
```

saisankargochhayat: Does this only help in sentry grouping? Or is it because it has a better performance?

fridex: Performance - the string is formatted based on log level.

PR #648, opened 2020-01-28
Extend parameters for Adviser

pacospace:
Depends-On: https://github.com/thoth-station/user-api/pull/758/files


thoth/common/openshift.py
```diff
@@ -1579,6 +1583,16 @@ def schedule_adviser(
             runtime_environment
         )
 
+        template_parameters["THOTH_ADVISER_METADATA"] = json.dumps(
```
fridex: This change should also land in `run_adviser`. Otherwise, parameters propagated to `run_adviser` method will not match to parameters provided by `schedule_adviser` method.

fridex:
regate

PR #663, opened 2020-02-05
Add method to retrieve image streams

pacospace:

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -469,6 +469,14 @@ class OpenShift:
         )
         return result
 
+    def get_image_streams(self, namespace: str, label_selector: str) -> Dict[str, Any]:
+        """Get all image streams in a namespace and select them by label."""
+        v1_image_streams = self.ocp_client.resources.get(api_version="image.openshift.io/v1", kind="ImageStream")
+        result: Dict[str, Any] = v1_image_streams.get(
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -469,6 +469,14 @@ def get_configmaps(self, namespace: str, label_selector: str) -> Dict[str, Any]:
         )
         return result
 
+    def get_image_streams(self, namespace: str, label_selector: str) -> Dict[str, Any]:
+        """Get all image streams in a namespace and select them by label."""
+        v1_image_streams = self.ocp_client.resources.get(api_version="v1", kind="ImageStream")
```
fridex: `api_version="image.openshift.io/v1"`

thoth/common/openshift.py
```diff
@@ -469,6 +469,14 @@ def get_configmaps(self, namespace: str, label_selector: str) -> Dict[str, Any]:
...
+        result: Dict[str, Any] = v1_image_streams.get(
+            label_selector=label_selector, namespace=namespace
+        )
+        return result
```
fridex: Will it retrieve a list? (IIRC there is an "items" key?)

pacospace: No a Dict[str, Any] as for configmaps.

PR #705, opened 2020-03-03
Generalize variables for Ceph storage for workflow

pacospace:

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -26,6 +26,8 @@ import random
 import urllib3
 import time
 
+from urllib.parse import urlparse
+
 from typing import Any
 from typing import Dict
 from typing import List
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -1683,6 +1687,22 @@ def schedule_adviser(
             },
         )
 
+    def _assign_workflow_parameters(self) -> Dict[str, Any]:
+        """Check and assign workflow parameters for different services."""
+        self.ceph_bucket_prefix = os.environ["THOTH_CEPH_BUCKET_PREFIX"]
+        self.ceph_bucket_name = os.environ["THOTH_CEPH_BUCKET"]
+        self.ceph_host = urlparse(os.environ["THOTH_S3_ENDPOINT_URL"]).netloc
+        self.deployment_name = os.environ["THOTH_DEPLOYMENT_NAME"]
```
fridex: OpenShift class does not have these properties, assign them directly to the dict retuned:

```
        return {
            "ceph_bucket_prefix": os.environ["THOTH_CEPH_BUCKET_PREFIX"],
     ...
```

thoth/common/openshift.py
```diff
@@ -1683,6 +1687,17 @@ def schedule_adviser(
             },
         )
 
+    def _assign_workflow_parameters(self) -> Dict[str, Any]:
```
fridex: We should move scheduling of inspection workflow manager to this OpeNshift class. It logically should be placed here, this method would be also reused in that case.

pacospace: I can do that sure, I already open PR for adviser workflow as well, only thing we need to solve thoth-common software stack :(

PR #713, opened 2020-03-11
Fixes for thamos advise

CermakM:
This commit fixes Workflow ID being reported by thamos instead of the adviser ID and it prevents thamos from failing to wait for an analysis if a Job has not yet been created.

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1517,7 +1517,7 @@ class OpenShift:
 
         workflow_parameters = self._assign_workflow_parameters_for_ceph()
 
-        _ =  self._schedule_workflow(
+        _ = self._schedule_workflow(
             workflow=self.workflow_manager.submit_adviser_workflow,
             parameters={
                 "template_parameters": template_parameters,
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1517,7 +1517,7 @@ class OpenShift:
 
         workflow_parameters = self._assign_workflow_parameters_for_ceph()
 
-        _ = self._schedule_workflow(
+        self._schedule_workflow(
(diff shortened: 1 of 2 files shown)


thoth/common/openshift.py
```diff
@@ -1679,14 +1679,16 @@ def schedule_adviser(
 
         workflow_parameters = self._assign_workflow_parameters()
 
-        return self._schedule_workflow(
+        _ =  self._schedule_workflow(
```
fridex: No need to do assignment - the return value will be simply discarded.

thoth/common/openshift.py
```diff
@@ -2608,36 +2610,64 @@ def get_job_status_count(
...
-            )
-        except OpenShiftNotFoundError as exc:
-            raise NotFoundException(
-                f"The given Workflow containing label {label_selector} could not be found"
-            ) from exc
+        """Get Workflow from a namespace, use one of name or label_selector to identify which one to get."""
+        if not any([name, label_selector]):
+            raise ValueError("One of `name` or `label_selector` has to be provided.")
+        if name and label_selector:
+            raise ValueError("Either `name` or `label_selector` has to be provided.")
```
fridex: You can move these checks to the if-elif-else down bellow.

thoth/common/openshift.py
CermakM: isn't it a better practice according to some linters not to ignore a return value?

fridex: Which ones? I don't see any reason why they would require to do that. A lot of return values of function calls are simply ignored.

CermakM: Not sure. Just know that I've seen a message like that somewhere :D I will get rid of that, not a problem :)

CermakM:
recheck

PR #714, opened 2020-03-11
Introduce Schedule inspection method in Openshift Class

pacospace:


thoth/common/workflows.py
```diff
@@ -303,7 +302,6 @@ def submit_inspection_workflow(
         template_parameters = template_parameters or {}
         workflow_parameters = workflow_parameters or {}
 
-        template_parameters["AMUN_INSPECTION_ID"] = inspection_id
```
CermakM: This is being used by Amun API to identify the Workflows: https://github.com/thoth-station/amun-api/blob/b08157e7883df1f7745e9eff5c8395acf1fe8ff7/amun/api_v1.py#L349.

pacospace: I didn't remove it, I just moved it on Openshift class side :)

CermakM: Ah cool hehe, it's somewhat hard to compare the diffs across the PRs :) Thanks!

PR #715, opened 2020-03-12
Modify logic of get_solver_names not to depend on Openshift template

pacospace:
Tested and CM already updated.

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -693,23 +693,25 @@ class OpenShift:
             raise ConfigurationError(
                 "Infra namespace is required in order to list solvers"
             )
-        label_selector = "template=solver"
 
         if not self.use_argo:
-            label_selector = "template=solver-workload-operator"
+            response = self.ocp_client.resources.get(
+                api_version="template.openshift.io/v1", kind="Template", name="templates"
+            ).get(namespace=self.infra_namespace, label_selector="template=solver-workload-operator")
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -693,23 +693,25 @@ def get_solver_names(self) -> List[str]:
...
-            "OpenShift response for getting solver template: %r", response.to_dict()
-        )
-        self._raise_on_invalid_response_size(response)
-        return [
-            obj["metadata"]["labels"]["solver-type"]
-            for obj in response.to_dict()["items"][0]["objects"]
-            if "solver-type" in obj["metadata"]["labels"]
-        ]
+        cm = self.get_configmap("thoth", namespace=self.infra_namespace)
+        solvers = [s.strip() for s in cm["data"]["solvers"].splitlines()]
```
fridex: Not to cause fatal crash if we don't have any installed/entries present:
```
        solvers = [s.strip() for s in cm["data"].get("solvers", "").splitlines()]
```

thoth/common/openshift.py
```diff
@@ -693,23 +693,25 @@ def get_solver_names(self) -> List[str]:
             raise ConfigurationError(
                 "Infra namespace is required in order to list solvers"
             )
-        label_selector = "template=solver"
 
         if not self.use_argo:
```
fridex: I think we can drop the template logic completely.

PR #716, opened 2020-03-12
Fix issue with returning None instead of workflow id

fridex:


CermakM:
Huh? I guess a rebase issue... This method used to return `adviser_id`. But returning the `_shedule_workflow` return value (which is the workflow name) should be ok as well as this should now map to the `adviser_id`.

PR #723, opened 2020-03-17
Add empty env template

KPostOffice:


.env.template
```diff
@@ -0,0 +1,27 @@
...
+OPENSHIFT_API_URL=
+THOTH_USE_ARGO=
+
+# Logging
+THOTH_LOG_SOLVER=
+THOTH_LOG_SOLVER__PYTHON=
+RSYSLOG_HOST=
+RSYSLOG_PORT=
+SENTRY_DSN=
+_LOGGING_ADJUSTMENT_CONF=
```
fridex: Not env var.

.env.template
```diff
@@ -0,0 +1,27 @@
...
+THOTH_CLUSTER_DEPLOYMENT=
```
fridex: This one is not used.

.env.template
```diff
@@ -0,0 +1,27 @@
...
+THOTH_MIDDLETIER_NAMESPACE=
+THOTH_BACKEND_NAMESPACE=
+THOTH_INFRA_NAMESPACE=
+KUBERNETES_API_URL=
+OPENSHIFT_API_URL=
+THOTH_USE_ARGO=
+
+# Logging
+THOTH_LOG_SOLVER=
+THOTH_LOG_SOLVER__PYTHON=
```
fridex: These two are not used within common code base - they are specific configuration when configuring logging.

PR #727, opened 2020-03-20
Add re_run_adviser_id metadata for qeb-hwt app

pacospace:

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1444,6 +1444,7 @@ class OpenShift:
         github_check_run_id: Optional[int] = None,
         github_installation_id: Optional[int] = None,
         github_base_repo_url: Optional[str] = None,
+        re_run_adviser_id: Optional[str] = None
     ) -> str:
         """Schedule an adviser run."""
         if not self.backend_namespace:
@@ -1492,6 +1493,7 @@ class OpenShift:
                 "github_base_repo_url": github_base_repo_url,
                 "origin": origin,
                 "is_s2i": is_s2i,
+                "re_run_adviser_id": re_run_adviser_id
             }
         )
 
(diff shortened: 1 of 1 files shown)


pacospace:
> Hm, how would the logic look like then?
> 
> Wouldn't be it easier to mark it in the database? In that case, the query for adviser runs that need to be re-run would already incorporate this flag in it and we don't need to iterate (download) results from ceph.

So we schedule the new adviser and we change that flag in the database after scheduling for the adviser run where we took the info?

fridex:
> > Hm, how would the logic look like then?
> > Wouldn't be it easier to mark it in the database? In that case, the query for adviser runs that need to be re-run would already incorporate this flag in it and we don't need to iterate (download) results from ceph.
> 
> So we schedule the new adviser and we change that flag in the database for the adviser run where we took the info, right after scheduling ?

I think so. Otherwise, we would need to go to ceph and load whole JSON results (more expensive than having a binary flag in the database that can be used during the query).

pacospace:
> > > Hm, how would the logic look like then?
> > > Wouldn't be it easier to mark it in the database? In that case, the query for adviser runs that need to be re-run would already incorporate this flag in it and we don't need to iterate (download) results from ceph.
> > 
> > 
> > So we schedule the new adviser and we change that flag in the database for the adviser run where we took the info, right after scheduling ?
> 
> I think so. Otherwise, we would need to go to ceph and load whole JSON results (more expensive than having a binary flag in the database that can be used in the query).

Modified as we discussed in the meeting. Thanks!! :beers:

pacospace:
recheck

fridex:
> * [thoth-pytest ](https://managesf.thoth-station.ninja/zuul/t/local/build/173619d96a614ca086f5712060f9a1f0) : FAILURE in 4m 17s

https://github.com/thoth-station/common/pull/728

PR #744, opened 2020-03-25
Remove warning produced by jsonformatter

fridex:
From OpenShift logs:

```
/opt/app-root/lib/python3.6/site-packages/jsonformatter/jsonformatter.py:76: UserWarning: Current Python version is below 3.7.0, the key's order of dict may be different from the definition, Please Use `OrderedDict` replace.
--
  | "Current Python version is below 3.7.0, the key's order of dict may be different from the definition, Please Use `OrderedDict` replace.", UserWarning)

```

### DIFF
--- a/thoth/common/logging.py
+++ b/thoth/common/logging.py
@@ -22,6 +22,7 @@ import sys
 import logging
 import socket
 import time
+from collections import OrderedDict
 from typing import Optional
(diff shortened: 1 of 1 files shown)


saisankargochhayat:
@fridex  looks good to be merged! :)

PR #759, opened 2020-04-01
Allow to set use_argo and initialize workflow_manager

CermakM:
- schedule_inspection warns about using argo workflows

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -146,14 +146,8 @@ class OpenShift:
                 "TLS verification when communicating with k8s/okd master is disabled"
             )
 
-        self.use_argo = use_argo or bool(int(os.getenv("THOTH_USE_ARGO", 0)))
-
-        self.workflow_manager = None
-        if self.use_argo:
-            _LOGGER.info("Using Argo Workflow to run jobs")
-            from .workflows import WorkflowManager
-
-            self.workflow_manager = WorkflowManager(openshift=self)
+        self._use_argo = use_argo or bool(int(os.getenv("THOTH_USE_ARGO", 0)))
+        self._workflow_manager = None
 
     @property
     def token(self) -> str:
@@ -168,6 +162,32 @@ class OpenShift:
                 ]["value"].split(" ")[1]
 
         return self._token
+
+    @property
+    def use_argo(self) -> bool:
+        """Return whether Argo Workflows are currently enabled."""
+        return self._use_argo
+
+    @use_argo.setter
+    def use_argo(self, enable: bool):
+        """Set whether Argo Workflows are enabled."""
+        if enable:
+            _LOGGER.debug("Argo Workflows enabled")
+        else:
+            _LOGGER.debug("Argo Workflows disabled")
+
+        self._use_argo = enable
+
+    @property
+    def workflow_manager(self):
+        """Return WorkflowManager instance.
+
+        This property lazily initializes the WorkflowManager.
+        """
+        if self._workflow_manager is None:
+            from .workflows import WorkflowManager
+            self._workflow_manager = WorkflowManager(openshift=self)
+        return self._workflow_manager
 
     @staticmethod
     def _set_env_var(template: Dict[str, Any], **env_var: str) -> None:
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -169,6 +169,27 @@ def token(self) -> str:
 
         return self._token
 
+    @property
```
fridex, line 172: Wouldn't it be easier to create a property out of `workflow_manager` and instantiate it once needed?

```python
    @property
    def workflow_manager(self) -> bool:
        if self._workflow_manager is None:
            self._workflow_manager = WorkflowManager(openshift=self)
         return self._workflow_manager
```

This way the following code will work as expected:

```python
os = OpenShift()
os.use_argo = True
os.schedule_foo(**kwargs)
os.use_argo = False
os.schedule_foo(**kwargs)
os.use_argo = True
os.schedule_foo(**kwargs)
os.use_argo = False
os.schedule_foo(**kwargs)
```

CermakM: How is the code not working as expected?

The workflow_manage can probably be a separate property, as well, but that would still require being able to set the `use_argo`, otherwise there would be an inconsistency between `use_argo=True` and `workflow_manager=None` and vice versa.

fridex: `schedule_thamos_workflow` has no legacy counterpart implementation part, so it will raise an exception when `workflow_manager` will be set to None. Also the complexity behind this patch - workflow manager could be instantiated lazily as needed and the logic behind `schedule_*` methods can decide what implementation should be used.

CermakM: Right, good point, so, first of all, we should perhaps raise an error instead of logging the warning for those methods which don't have the legacy counterpart.

And about the complexity, if as the OpenShift is already initialized, importing the Workflow manager is just importing a class. But if we want to do that only once, then `use_argo=False` perhaps shouldn't set the `workflow_manager` to `None`, then?

fridex: > Right, good point, so, first of all, we should perhaps raise an error instead of logging the warning for those methods which don't have the legacy counterpart.

That will prevent from scheduling thamos workflow in all deployments that have `THOTH_USE_ARGO=0` (now stage and prod for demo purposes).


> And about the complexity, if as the OpenShift is already initialized, importing the Workflow manager is just importing a class. But if we want to do that only once, then `use_argo=False` perhaps shouldn't set the `workflow_manager` to `None`, then?

Yes, we can instantiate the workflow manager all the time. Note we are in the middle of the transition to argo, once done, the legacy implementation will go away completely.

CermakM: Uhm, I guess it is good idea to raise as it would collide with the rest of the application and would be harder to set up. You're right, I'll use the property initialization you suggested. I can still leave the `use_argo` method, can't I tho?

fridex: > Uhm, I guess it is good idea to raise as it would collide with the rest of the application and would be harder to set up. 

We want to run thamos workflows (github app specific workflows) that are implemented only in argo workflows. We also want to have the ability to switch between the legacy implementation of workflows (adviser, solver, ...) that use workload-operator or argo workflows. Not all workflows are ported to argo and not all workflows are available for the legacy workload-operator (thamos workflow). As said, we are in the middle of transition so I don't see any straightforward way how to test/run argo workflows besides legacy workload-operator implementation until we fully port thoth to argo without losing some of its functionality.

> You're right, I'll use the property initialization you suggested. I can still leave the `use_argo` method, can't I tho?

If I follow correctly, `use_argo` will be always set to `True`/`False`.

CermakM: Yes, correct ... sorry didn't see your comments ( the window does not refresh automatically :/ )

Implemented the changes, please, take a look :) I did remove the `workflow_manager=None` when setting the `use_argo` to False as it seems somewhat unnecessary as it would be initialized anyway by the property accessor.

CermakM: EDIT: I meant "guess it is NOT good idea" ... man, some race conditions in this chat :D

CermakM: > If I follow correctly, use_argo will be always set to True/False.

Yes, so now it doesn't make any sense to have them as properties anymore, right, as it overcomplicates things and it can simply be an attribute. Now I follow. :D

CermakM: We can still leave them there tho for the documentation purposes so that users know how to turn on/off this behaviour.

CermakM:
Rebased and squashed for clarity.

fridex:
recheck

PR #770, opened 2020-04-08
Propagate requests and limits for inspection run and build

fridex:


fridex:
I'll update the inspection templates accordingly.

PR #771, opened 2020-04-08
Remove parallelism and allowed failures configuration

fridex:
Parallelism will be handled by Argo configuration based on cluster resources.
Allowed failures will not be taken into account with argo tasks.


fridex:
I'll update the templates accordingly.

PR #774, opened 2020-04-08
Remove assignment of the Dockerfile

fridex:
It's assigned to Argo template. This removes huge warning printed to logs.


fridex:
regate

PR #791, opened 2020-04-27
Moved json serialization

saisankargochhayat:
Moved json serialization to thoth-common.


fridex:
the parameter type should be also changed from `str` to `Dict[str, Any]`.

PR #807, opened 2020-05-13
add is_kebechet to metadata

KPostOffice:
Add metadata so that we can use it to add new task to adviser workflow


KPostOffice:
See: https://github.com/thoth-station/common/pull/811

PR #811, opened 2020-05-15
Add source type for Thoth adviser integrations

pacospace:

### DIFF
--- a/thoth/common/enums.py
+++ b/thoth/common/enums.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+# thoth-common
+# Copyright(C) 2020 Francesco Murdaca
+#
+# This program is free software: you can redistribute it and / or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Enum types used for Thoth adviser integrations."""
+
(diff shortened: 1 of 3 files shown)


thoth/common/openshift.py
```diff
@@ -1479,6 +1481,7 @@ def schedule_adviser(
         github_installation_id: Optional[int] = None,
         github_base_repo_url: Optional[str] = None,
         re_run_adviser_id: Optional[str] = None,
+        source_type: Optional[ThothAdviserIntegrationEnum] = None,
```
KPostOffice, line 1484: We can remove `is_s2i` here and below

pacospace: I can add a TODO so @fridex can take it, I don't know how it can affect the thoth-s2i image if we remove that parameter right now. ack?

KPostOffice: ack. Good call :+1:

thoth/common/enums.py
```diff
@@ -0,0 +1,30 @@
...
+
+"""Enum types used for Thoth adviser integrations."""
+
+from enum import Enum
+
+
+class ThothAdviserIntegrationEnum(Enum):
+    """Class for Thoth Adviser integrations."""
+
+    cli = "cli"
```
fridex, line 26: You can use `auto` to reduce overhead with string comparision and be more enum-like:

https://docs.python.org/3/library/enum.html#enum.auto

Also, enums tend to be upper case.

thoth/common/openshift.py
```diff
@@ -1469,7 +1471,7 @@ def schedule_adviser(
         runtime_environment: Optional[Dict[Any, Any]] = None,
         library_usage: Optional[Dict[Any, Any]] = None,
         origin: Optional[str] = None,
-        is_s2i: Optional[bool] = None,
+        is_s2i: Optional[bool] = None,  # TODO: Remove is_s2i parameter everywhere (substitute: source_type)
```
fridex, line 1474: > Remove is_s2i parameter everywhere (substitute: source_type)

Can we do it with this change so the release does not carry technical depth?

goern:
recheck

PR #816, opened 2020-05-20
Create exception

pacospace:


thoth/common/openshift.py
```diff
@@ -1460,6 +1460,14 @@ def get_build_dependencies_template(self) -> Dict[str, Any]:
 
         return self._get_template("template=build-dependencies")
 
+    @staticmethod
+    def _verify_thoth_integration(source_type: str):
+        """Verify if Thoth integration exists."""
+        integrations = [i.name for i in ThothAdviserIntegrationEnum]
+
+        if source_type not in integrations:
```
fridex, line 1468: Will ``None`` (no value provided) work here?

```python
        if source_type not in integrations and source_type is not None
```

fridex: btw enum has ``__members__``

PR #817, opened 2020-05-20
New check method for Qeb-HWt inputs

pacospace:

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1465,8 +1465,9 @@ class OpenShift:
         """Verify if Thoth integration exists."""
         if source_type not in ThothAdviserIntegrationEnum.__members__ and source_type is not None:
             raise NotKnownThothIntegration(
-        f"This integration {source_type} is not provided in Thoth: {ThothAdviserIntegrationEnum.__members__.keys()}"
-    )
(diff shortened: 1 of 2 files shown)


thoth/common/openshift.py
```diff
@@ -1461,12 +1461,26 @@ def get_build_dependencies_template(self) -> Dict[str, Any]:
         return self._get_template("template=build-dependencies")
 
     @staticmethod
-    def _verify_thoth_integration(source_type: str):
+    def _verify_thoth_integration(source_type: Optional[str]):
         """Verify if Thoth integration exists."""
-        integrations = [i.name for i in ThothAdviserIntegrationEnum]
+        if source_type not in ThothAdviserIntegrationEnum.__members__ and source_type is not None:
```
KPostOffice, line 1466: this should be `.__members__.keys()`

KPostOffice: Oh wait, I forgot about behavior of in with dictionaries, this is actually good

pacospace: The dict read only the keys if you don't state differently :) 

for `k in dict_name` or  `for k in dict_name.keys()` should be equivalent

PR #833, opened 2020-05-26
Use precommit

KPostOffice:
Many changes using new pre-commit hooks.  Added checks where `None` checks were missing, added type annotations to functions. Renamed tests. Black and pyflakes styling.


harshad16:
i feel like changes in  .pre-commit-config.yaml should be shifted to another pr as its only thing needs to be reviewed and rest as it due to the pre-commit requirements can be review more leniently. 
WDYT?

goern:
> i feel like changes in .pre-commit-config.yaml should be shifted to another pr as its only thing needs to be reviewed and rest as it due to the pre-commit requirements can be review more leniently.
> WDYT?

Ja, separating would be great.

goern:
If we run the result thru black, what will happen? will it complain and change back?

KPostOffice:
@goern this is using black formatting if you run it through again nothing should change

KPostOffice:
recheck

KPostOffice:
I don't think this will be automatically merged because it expects that it will pass `coala` which is impossible based on the new constraints from `precommit`

fridex:
> I don't think this will be automatically merged because it expects that it will pass `coala` which is impossible based on the new constraints from `precommit`

I think we can keep coala running and configure its pycodestyle bear so that it accepts code linted with black (ignoring restrictions like space before the colon).

fridex:
Or start using flake8 for checks as [done here](https://github.com/pypa/packaging/blob/59e14881910bf836c43b7d319ea08ab321ccfb39/.pre-commit-config.yaml#L29-L32).

You can keep coafile as it also verifies other files, not only Python code (e.g. YAML files).

KPostOffice:
@fridex This is using flake8 and black

PR #837, opened 2020-05-29
Introudce method for verifying Kebechet inputs

pacospace:

### DIFF
--- a/thoth/common/exceptions.py
+++ b/thoth/common/exceptions.py
@@ -41,5 +41,9 @@ class QebHwtInputsMissing(ThothCommonException):
     """An exception raised if there are inputs missing for QebHwt Github App."""
 
 
+class KebechetInputsMissing(ThothCommonException):
+    """An exception raised if there are inputs missing for Kebechet."""
+
+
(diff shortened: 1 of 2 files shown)


pacospace:
> Do we want to have other parameters similar to the one we have from Qeb-hwt? I am not sure which all we can extract, but we do receive the whole webhook if we want to keep this parameters for later? Or would these parameters not be important, as Kebechet uses O_auth token to interact with the repositories.
> If we just want the source_type, this LGTM

as far as I see from kebechet thoth integration it requires origin at least right?

saisankargochhayat:
> > Do we want to have other parameters similar to the one we have from Qeb-hwt? I am not sure which all we can extract, but we do receive the whole webhook if we want to keep this parameters for later? Or would these parameters not be important, as Kebechet uses O_auth token to interact with the repositories.
> > If we just want the source_type, this LGTM
> 
> as far as I see from kebechet thoth integration it requires origin at least right? @saisankargochhayat

Yes just the origin

saisankargochhayat:
recheck

PR #844, opened 2020-06-04
Bump black version to 19.10b0

fridex:


goern:
Uhhh, why is that?

fridex:
> Uhhh, why is that?

I'm getting different results when I format code locally from the ones from CI.

goern:
> > Uhhh, why is that?
> 
> I'm getting different results when I format code locally from the ones from CI.

so we set a policy that we only use black 19.10b0 with all repositories and locally?

fridex:
> > > Uhhh, why is that?
> > 
> > 
> > I'm getting different results when I format code locally from the ones from CI.
> 
> so we set a policy that we only use black 19.10b0 with all repositories and locally?

`19.10b0` is the most recent version now. It might be good idea to keep up with the most recent releases in this case.

PR #879, opened 2020-06-23
some reformatting

goern:
this will keep pre-commit happy

Signed-off-by: Christoph Görn <goern@redhat.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ build/
 .coverage
 coverage.xml
 .env
+.vscode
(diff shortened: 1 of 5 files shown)


fridex:
```
thoth/common/openshift.py:1292:16: error: Incompatible return value type (got "Optional[str]", expected "str")  [return-value]
```


```
black....................................................................Failed
```

Maybe an update on your host will be needed:

https://github.com/thoth-station/common/blob/7280623c609a2eda4635c864c386f5f7592246be/.pre-commit-config.yaml#L45-L46

PR #901, opened 2020-07-02
Use singular

pacospace:


fridex:
This will affect other components relying on the method names CC @xtuchyna

PR #907, opened 2020-07-07
Introduce reverse solver workflow

fridex:
Let's create whole revsolver workflow, it will be easier to handle.


goern:
@fridex is there a PR containing the required templates?

PR #909, opened 2020-07-15
Make methods static

pacospace:


thoth/common/openshift.py
```diff
@@ -224,8 +224,10 @@ def parse_python_solver_name(cls, solver_name: str) -> Dict[str, Any]:
             "python_version": python_version,
         }
 
+    @classmethod
```
saisankargochhayat, line 227: Why are we changing to a classmethod?

PR #918, opened 2020-07-21
Provide cached Python package version tuple

fridex:


thoth/common/config/runtime_environment.py
```diff
@@ -44,6 +45,7 @@ class RuntimeEnvironment:
     cuda_version = attr.ib(type=str, default=None)
     name = attr.ib(type=str, default=None)
     platform = attr.ib(type=str, default=None)
+    _python_version_tuple = attr.ib(type=Optional[Tuple[int, int]], default=None)
```
KPostOffice, line 48: `init=False`?

goern:
ping?!

PR #919, opened 2020-07-23
Added workflow status wrapper

saisankargochhayat:
Added a wrapper around `get_workflow_status`. 
Related to https://github.com/thoth-station/user-api/issues/965

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1608,6 +1608,29 @@ class OpenShift:
 
         return wf
 
+    def get_workflow_status_report(
+        self,
+        workflow_id: str,
+        label_selector: Optional[str] = None,
+        namespace: Optional[str] = None,
+    ) -> Dict[str, Optional[str]]:
+        """Get workflow status report, derived from workflow status."""
+        try:
+            wf_status = self.get_workflow_status(
+                name=workflow_id, label_selector=label_selector, namespace=namespace
+            )
+            status = {
+                "finished_at": wf_status.get("finishedAt"),
+                "reason": None,
+                "started_at": wf_status.get("startedAt"),
+                "state": wf_status.get("phase", "pending").lower(),
+            }
+        except NotFoundException as exc:
+            raise NotFoundException(
+                f"Requested status for analysis {workflow_id!r} was not found"
+            ) from exc
+        return status
+
     def get_workflow_status(
         self,
         name: Optional[str] = None,


thoth/common/openshift.py
```diff
@@ -1608,6 +1609,43 @@ def get_workflow(
...
 
+    def get_workflow_status_report(
+        self,
+        workflow_id: Optional[str] = None,
+        label_selector: Optional[str] = None,
+        namespace: Optional[str] = None,
+    ) -> Tuple[Dict[str, Any], int]:
+        """Get workflow status report, derived from workflow status."""
+        if workflow_id is None:
+            return {"error": "No analysis id provided"}, 400
```
KPostOffice, line 1620: I wouldn't use the term `analysis` here as this can be any workflow

thoth/common/openshift.py
```diff
@@ -1608,6 +1609,43 @@ def get_workflow(
...
+            wf_status = self.get_workflow_status(
+                name=workflow_id, label_selector=label_selector, namespace=namespace
+            )
+            status = {}
+            # the Job has not started (yet) but the Workflow has been submitted
+            # if the Workflow fails, the status will contain the finished
+            # time of the workflow
+            status["status"], code = (
+                {
+                    "finished_at": wf_status.get("finishedAt"),
```
KPostOffice, line 1631: Will these always be available not matter what phase the workflow is in?

thoth/common/openshift.py
```diff
@@ -1608,6 +1609,43 @@ def get_workflow(
...
+            )
+        except NotFoundException:
+            return (
+                {
+                    "error": f"Requested status for analysis {workflow_id!r} was not found"
+                },
+                404,
+            )
+        except Exception:
+            _LOGGER.error(status["error"])
```
KPostOffice, line 1646: What should the function return in this case?  It will get to the next line and `status` and `code` have not been set.

thoth/common/openshift.py
```diff
@@ -1608,6 +1609,43 @@ def get_workflow(
...
+                    "state": wf_status.get("phase", "Pending"),
+                },
+                200,
+            )
+        except NotFoundException:
+            return (
+                {
+                    "error": f"Requested status for analysis {workflow_id!r} was not found"
+                },
+                404,
```
KPostOffice, line 1643: I think `400` (bad request) is more apt.

thoth/common/openshift.py
```diff
@@ -1608,6 +1609,43 @@ def get_workflow(
...
+                name=workflow_id, label_selector=label_selector, namespace=namespace
+            )
+            status = {}
+            # the Job has not started (yet) but the Workflow has been submitted
+            # if the Workflow fails, the status will contain the finished
+            # time of the workflow
+            status["status"], code = (
+                {
+                    "finished_at": wf_status.get("finishedAt"),
+                    "reason": None,
```
KPostOffice, line 1632: Why is `reason` included in the status if it will always be set to None.

thoth/common/openshift.py
saisankargochhayat: I think we should raise an exception here, WDYT?

thoth/common/openshift.py
saisankargochhayat: This is to make it compatible to the existing [status report](ttps://github.com/thoth-station/user-api/blob/c1b80baae49b90556534f3c5de3c5e47a0c8d530/thoth/user_api/api_v1.py#L360).

thoth/common/openshift.py
saisankargochhayat: 400 is used when the request is a malformed type like passing an int instead, but 404 seems the right choice considering the format is correct, but resource was not found.

thoth/common/openshift.py
saisankargochhayat: `finished_at` would be `None`, when the workflow is in a running state 
For ex - `{'status': {'finished_at': None, 'reason': None, 'started_at': '2020-07-23T20:59:28Z', 'state': 'Running'}} 200`

thoth/common/openshift.py
fridex: Please move HTTP status codes out of this module. They can be added on API services (user-api/management-api/...), this module should not capture them. For example, these functions can be used also on metrics exporter to export info about running/pending/... workflows and HTTP status code is not relevant there.

thoth/common/openshift.py
fridex: Do we want to add some info here (e.g. why a workflow failed) or do we want to completely discard this item?

thoth/common/openshift.py
fridex: You don't need to do generic except block here. If any exception occurs, just let it be propagated.

thoth/common/openshift.py
```diff
@@ -1608,6 +1609,43 @@ def get_workflow(
...
+                    "started_at": wf_status.get("startedAt"),
+                    "state": wf_status.get("phase", "Pending"),
+                },
+                200,
+            )
+        except NotFoundException:
+            return (
```
fridex, line 1639:
```python
       except NotFoundException as exc:
            raise NotFoundException(f"Requested status for analysis {workflow_id!r} was not found") from exc
```

thoth/common/openshift.py
saisankargochhayat: Will remove it :+1:

thoth/common/openshift.py
saisankargochhayat: @fridex Maybe if we comb through all the nodes, we can find info like which workflows failed or something?
If the particular step has finished and failed, it has a [message key](https://gist.github.com/saisankargochhayat/6ed94ee54f8221053fd2f1a29f560e63#file-adviser_status-json-L116), we could iterate through nodes to see if such a message exists? I am not sure if the message is same for all. 
This is what I found from the documentation - https://argoproj.github.io/argo/swagger/#ioargoprojworkflowv1alpha1workflowstatus
However all the keys aren't there in the dict we are receiving, like `message` in the root level.

fridex: This might be tricky to implement for users of Thoth who do not see/don't want to know details about workflows. As workflows can fail in many places (and still be successful based on the configuration like skips on an error), let's keep it as None for now for simplicity (or some generic message). Some of these messages are not useful for end-users, like `failed with exit code 1` or `No more retries left`.

We can check what errors we observe and if we want to populate them to users (with a descriptive error message).

WDYT?

saisankargochhayat: Makes sense, let's keep it none for now, however let's have key in-case we want to pass more information in the future.

saisankargochhayat:
@fridex should we merge this?

thoth/common/openshift.py
```diff
@@ -1608,6 +1608,35 @@ def get_workflow(
 
         return wf
 
+    def get_workflow_status_report(
+        self,
+        workflow_id: Optional[str] = None,
```
fridex, line 1613: It is not optional.

thoth/common/openshift.py
```diff
@@ -1608,6 +1608,35 @@ def get_workflow(
...
+        label_selector: Optional[str] = None,
+        namespace: Optional[str] = None,
+    ) -> Dict[str, Any]:
+        """Get workflow status report, derived from workflow status."""
+        if workflow_id is None:
+            return {"error": "No workflow id provided"}
```
fridex, line 1619: workflow_id should not be optional, you can discard this check

thoth/common/openshift.py
```diff
@@ -1608,6 +1608,35 @@ def get_workflow(
...
+        try:
+            wf_status = self.get_workflow_status(
+                name=workflow_id, label_selector=label_selector, namespace=namespace
+            )
+            status = {}
+            # the Job has not started (yet) but the Workflow has been submitted
```
fridex, line 1625: Is this comment still relevant?

thoth/common/openshift.py
```diff
@@ -1608,6 +1608,35 @@ def get_workflow(
...
+            # if the Workflow fails, the status will contain the finished
+            # time of the workflow
+            status["status"] = {
```
fridex, line 1628: Can we return directly status? I think making a nested dict does not make sense in this case.

saisankargochhayat: I agree, will change that

thoth/common/openshift.py
saisankargochhayat: Removed it

saisankargochhayat:
@fridex thanks for the review, PTAL :)

PR #932, opened 2020-08-07
set source_type_enum None when soure_type not set

harshad16:
set source_type_enum None when soure_type not set
Fixes: #931 
Signed-off-by: Harshad Reddy Nalla <hnalla@redhat.com>

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1175,11 +1175,9 @@ class OpenShift:
 
         if source_type is not None:
             self._verify_thoth_integration(source_type=source_type)
-            source_type_enum = (
-                getattr(ThothAdviserIntegrationEnum, source_type)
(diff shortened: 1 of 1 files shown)


saisankargochhayat:
> Maybe I'm missing something. What's the change here?

I think the source_type would be not declared - https://github.com/thoth-station/common/blob/d7ed5635884c21cbbea9cfd1f2a56a0cdb803242/thoth/common/openshift.py#L1182

fridex:
> > Maybe I'm missing something. What's the change here?
> 
> I think the source_type would be not declared -
> 
> https://github.com/thoth-station/common/blob/d7ed5635884c21cbbea9cfd1f2a56a0cdb803242/thoth/common/openshift.py#L1182

I missed the indentation change :)

PR #948, opened 2020-08-26
add optional limit for workflows

KPostOffice:
https://github.com/thoth-station/investigator/issues/155

### DIFF
--- a/thoth/common/workflows.py
+++ b/thoth/common/workflows.py
@@ -460,9 +460,13 @@ class WorkflowManager:
 
         return wf.name
 
-    def get_total_workflows(self, workflow_namespace: str) -> int:
+    def get_pending_workflows(self, workflow_namespace: str) -> int:
         """Get the total number of workflows in a given namespace."""
-        return len(self.api.list_namespaced_workflows(workflow_namespace).items)
+        return len(
+            self.api.list_namespaced_workflows(
+                workflow_namespace, labels="workflows.argoproj.io/phase=Pending"
+            ).items
+        )
 
     def submit_workflow_from_template(
         self,
--- a/thoth/common/workflows.py
+++ b/thoth/common/workflows.py
@@ -460,6 +460,9 @@ class WorkflowManager:
 
         return wf.name
 
+    def _get_total_workflows(self, workflow_namespace: str) -> int:
+        return len(self.api.list_namespaced_workflows(workflow_namespace).items)
+
     def submit_workflow_from_template(
         self,
         namespace: str,
@@ -468,6 +471,7 @@ class WorkflowManager:
         template_parameters: Optional[Dict[str, str]] = None,
         workflow_parameters: Optional[Dict[str, Any]] = None,
         workflow_namespace: Optional[str] = None,
+        workflow_limit: Optional[int] = None,
     ) -> Union[str, None]:
         """Retrieve and Submit Workflow from an OpenShift template.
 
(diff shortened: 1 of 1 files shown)


thoth/common/workflows.py
```diff
@@ -480,7 +484,14 @@ def submit_workflow_from_template(
         :param template_parameters: parameters for the template
         :param workflow_parameters: parameters for the workflow
         :param workflow_namespace: namespace to submit the workflow to
+        :param workflow_limit: limit number of workflows currently in memory for workflowController
         """
+        if workflow_limit is not None:
+            if workflow_namespace is None:
+                raise ValueError("Namespace must be set when passing a limit.")
+            while self._get_total_workflows(workflow_namespace) > workflow_limit:
+                _LOGGER.debug("Waiting for number of workflows to drop below limit.")
```
fridex, line 493: Can this be done on the library user side? It feels not that nice to do it on the library side.

Also, the active wait will issue a lot of requests, some sleeping would be appropriate in this case.

KPostOffice: Yeah, a sleep would be good here.  I thought it would be nice to implement here, otherwise we will be duplicating the functionality a lot.

thoth/common/workflows.py
```diff
@@ -166,6 +167,8 @@ def __deserialize(cls, body: Dict[str, str], *, validate: bool) -> "Workflow":
 class WorkflowManager:
     """Argo Workflow manager."""
 
+    LIMIT_WAIT_TIME = 2.5  # time to wait between limit checking of argo api
```
fridex, line 170: Let's have it configurable via env var so we can adjust this in deployment without shipping a new component version.

thoth/common/workflows.py
fridex: > otherwise we will be duplicating the functionality a lot.

How many components would call this method?

KPostOffice: It would be for any component which is a `faust` consumer. Which includes at least investigator, package-update-job, and package-release

fridex: > package-update-job, and package-release

I think all the components should produce messages which will be handled by investigator - so I think only investigator will schedule workflows.

KPostOffice: Okay. I'll create an issue in investigator indicating this so that we know that we should be moving the logic over there.

KPostOffice:
I'll change this to just be a function that can be called rather than embedding it in the workflow scheduling.  The wait time will be an argument which can be passed to investigator

goern:
ping

thoth/common/workflows.py
```diff
@@ -460,6 +460,10 @@ def submit_workflow(
 
         return wf.name
 
+    def get_total_workflows(self, workflow_namespace: str) -> int:
+        """Get the total number of workflows in a given namespace."""
+        return len(self.api.list_namespaced_workflows(workflow_namespace).items)
```
fridex, line 465: Do we want to have just "pending" workflows?

harshad16:
all test are successfully completed , please neglect the pending thoth-ci checks as that service is stop, so it would stay at pending.
however it wouldn't hinder with the merging.

thoth/common/workflows.py
KPostOffice: I don't think the API supports this

fridex: > I don't think the API supports this

Can we do it programmatically on the library side then?

fridex (approved):
++

:+1:

PR #963, opened 2020-09-01
Hide default predictor name

fridex:


fridex:
So the default behavior works as expected:

https://github.com/thoth-station/management-api/blob/cd2167b1734126f207b4397895cfd560941c101a/thoth/management_api/api_v1.py#L206

pacospace (approved):

PR #966, opened 2020-09-07
Do not propagate private attribute in to_doct() method

fridex:

- [x] No


pacospace (approved):
/lgtm

PR #971, opened 2020-09-10
Provide a method for obtaining a link to justification description

fridex:

- [x] No

```python
>>> from thoth.common import get_justification_link as jl
>>> jl("oom")
'https://thoth-station.ninja/justifications/oom.html'
```

### DIFF
--- a/thoth/common/helpers.py
+++ b/thoth/common/helpers.py
@@ -169,3 +169,8 @@ class Lazy(object):
         value = self._calculate(obj)
         setattr(obj, self._calculate.__name__, value)
         return value
+
+
+def get_justification_link(identifier: str) -> str:
+    """Construct a link to a detailed justification document."""
+    return f"https://thoth-station.ninja/justifications/{identifier}.html"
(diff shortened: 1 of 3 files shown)


thoth/common/helpers.py
```diff
@@ -169,3 +169,8 @@ def __get__(self, obj: object, _: Any = None) -> Any:
         value = self._calculate(obj)
         setattr(obj, self._calculate.__name__, value)
         return value
+
+
+def get_justification_link(identifier: str) -> str:
+    """Construct a link to a detailed justification document."""
+    return f"https://thoth-station.ninja/justifications/{identifier}.html"
```
pacospace, line 176: don't we want this to be an env variable as maybe we are going to change something in the future?

pacospace (commented):
One small comment, otherwise very nice!!

thoth/common/helpers.py
fridex: Good idea, done in f4287ea

pacospace (approved):
:100 Thanks!

PR #982, opened 2020-09-11
allow job_id to be passed so we can return it

KPostOffice:
We need all workflows to be able to accept job_id because we are sending requests through kafka rather than directly through common.Openshift


fridex (approved):
:+1:

Issue #986, opened 2020-09-15
Adviser fails to run as wrong predictor config is supplied


fridex:
**Describe the bug**

```
2020-09-15 05:55:08,140   1 INFO     thoth.adviser:183: Version: 0.16.0
2020-09-15 05:55:08,247   1 CRITICAL root:102: Traceback (most recent call last):
  File "thoth-adviser", line 762, in <module>
    __name__ == "__main__" and cli()
  File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/app-root/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/opt/app-root/lib/python3.6/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "thoth-adviser", line 463, in advise
    predictor_instance = predictor_class(**predictor_kwargs, keep_history=plot is not None)
TypeError: type object argument after ** must be a mapping, not str

Sentry is attempting to send 1 pending error messages
Waiting up to 2 seconds
Press Ctrl-C to quit
```

**To Reproduce**
1. Schedule adviser in the cluster
2. See error

**Expected behavior**

The adviser should run.

PR #987, opened 2020-09-15
Fix predictor config propagation when scheduling adviser

fridex:

Fixes: https://github.com/thoth-station/common/issues/986

- [x] No


fridex:
Merging as hotfix as stage is affected by this.

PR #1007, opened 2020-10-05
get workflow node status

bissenbay:
Related-To: https://github.com/thoth-station/user-api/issues/1092

### DIFF
--- a/thoth/common/logging.py
+++ b/thoth/common/logging.py
@@ -131,7 +131,7 @@ def _logging_adjust() -> None:
 
 def _get_sentry_integrations() -> List[object]:
     """Get integrations for Sentry based on installed packages."""
-    integrations = []
+    integrations = []  # type: List[Any]
     try:
         import flask  # noqa
     except ImportError:
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -377,20 +377,34 @@ class OpenShift:
 
         return response.text
 
-    def get_workflow_node_log(
+    def get_workflow_pod_name(
         self, node_name: str, workflow_id: str, namespace: str
-    ) -> Optional[str]:
-        """Get log from a task/node in a workflow."""
+    ) -> str:
+        """Get pod name where task is being executed."""
         workflow = self.get_workflow(workflow_id, namespace=namespace)
         nodes = workflow.get("status", {}).get("nodes", {})
 
         for pod_name, node_info in nodes.items():
             if node_info["displayName"] == node_name:
-                return self.get_pod_log(pod_name, namespace=namespace, container="main")
+                return pod_name  # type: ignore
 
         raise NotFoundException(
-            f"No node {node_name!r} in workflow {workflow_id!r} in namespace {namespace!r} found"
+            f"No node named {node_name!r} found in workflow {workflow_id!r} in namespace {namespace!r}"
         )
+
+    def get_workflow_node_log(
+        self, node_name: str, workflow_id: str, namespace: str
+    ) -> Optional[str]:
+        """Get log from a task/node in a workflow."""
+        pod_name = self.get_workflow_pod_name(node_name, workflow_id, namespace)
+        return self.get_pod_log(pod_name, namespace=namespace, container="main")
+
+    def get_workflow_node_status(
+        self, node_name: str, workflow_id: str, namespace: str
+    ) -> Dict[str, Any]:
+        """Get status from a task/node in a workflow."""
+        pod_id = self.get_workflow_pod_name(node_name, workflow_id, namespace)
+        return self.get_pod_status(pod_id, namespace)
 
     def get_build(self, build_id: str, namespace: str) -> Dict[str, Any]:
         """Get a build in the given namespace."""


bissenbay:
@fridex I found that there is a method called `get_pod_status` at medthod https://github.com/thoth-station/common/blob/fd1229305a1dcaabe51eecbb290887bd1ccdf7d5/thoth/common/openshift.py#L487, I wonder if I can use it inside `get_workflow_node_status` to get the pod status or should we proceed with the one `get_pod_status` method which I introduced in this PR, I should probably just change the name of the function.

fridex:
> @fridex I found that there is a method called `get_pod_status` at medthod
> 
> https://github.com/thoth-station/common/blob/fd1229305a1dcaabe51eecbb290887bd1ccdf7d5/thoth/common/openshift.py#L487
> 
> , I wonder if I can use it inside `get_workflow_node_status` to get the pod status or should we proceed with the one `get_pod_status` method which I introduced in this PR, I should probably just change the name of the function.


You can refactor this part into a separate function:

```python
        workflow = self.get_workflow(workflow_id, namespace=namespace)
        nodes = workflow.get("status", {}).get("nodes", {})

        for pod_name, node_info in nodes.items():
            if node_info["displayName"] == node_name:
                # TODO: return self.get_pod_log(pod_name, namespace=namespace, container="main")
```

Taken from:

https://github.com/thoth-station/common/blob/fd1229305a1dcaabe51eecbb290887bd1ccdf7d5/thoth/common/openshift.py#L380-L383

bissenbay:
@fridex I tried to refactor common part for `get_workflow_node_log` & `get_workflow_node_status`. And also I found there is a method `get_pod_status`, what if we use that as I did in last commit instead of creating another function for getting pod status. And also just what to mention that `pod_name` and `pod_id` should be the same from my observation, but please let me know if I'm wrong. WDYT?

fridex (commented):
Looks good to me :100: Haven't check it, but it looks like it should work :+1:

fridex:
> @fridex I tried to refactor common part for `get_workflow_node_log` & `get_workflow_node_status`. And also I found there is a method `get_pod_status`, what if we use that as I did in last commit instead of creating another function for getting pod status.

Definitely, makes sense :+1: 

> And also just what to mention that `pod_name` and `pod_id` should be the same from my observation, but please let me know if I'm wrong. WDYT?

Yes, it's inconsistent in sources. :disappointed:

thoth/common/openshift.py
```diff
@@ -377,21 +377,35 @@ def get_pod_log(
...
     ) -> Optional[str]:
-        """Get log from a task/node in a workflow."""
+        """Get pod name where task is being executed."""
         workflow = self.get_workflow(workflow_id, namespace=namespace)
         nodes = workflow.get("status", {}).get("nodes", {})
 
         for pod_name, node_info in nodes.items():
             if node_info["displayName"] == node_name:
-                return self.get_pod_log(pod_name, namespace=namespace, container="main")
+                return pod_name
```
bissenbay, line 389: @fridex I'm getting a lot of pre-commit errors:
```
thoth/common/openshift.py:389:17: error: Returning Any from function declared to return "Optional[str]"  [no-any-return]
thoth/common/openshift.py:400:33: error: Argument 1 to "get_pod_log" of "OpenShift" has incompatible type "Optional[str]"; expected "str"  [arg-type]
thoth/common/openshift.py:407:16: error: Incompatible return value type (got "Dict[str, Any]", expected "Optional[str]")  [return-value]
thoth/common/openshift.py:407:36: error: Argument 1 to "get_pod_status" of "OpenShift" has incompatible type "Optional[str]"; expected "str"  [arg-type]
Found 4 errors in 1 file (checked 13 source files)
```
So, I tried returning `None` and moving 
```
raise NotFoundException(
     f"No node {node_name!r} in workflow {workflow_id!r} in namespace {namespace!r} found"
)
```
to `get_workflow_node_log` and `get_workflow_node_status`, but still not working, so I don't get why `pre-commit` is unhappy :smile: ?

fridex:
```
thoth/common/openshift.py:389:17: error: Returning Any from function declared to return "Optional[str]"  [no-any-return]
```

You are iterating over nodes but the type of `pod_name` is not known (hence type `Any`).

```
thoth/common/openshift.py:400:33: error: Argument 1 to "get_pod_log" of "OpenShift" has incompatible type "Optional[str]"; expected "str"  [arg-type]
thoth/common/openshift.py:407:16: error: Incompatible return value type (got "Dict[str, Any]", expected "Optional[str]")  [return-value]
thoth/common/openshift.py:407:36: error: Argument 1 to "get_pod_status" of "OpenShift" has incompatible type "Optional[str]"; expected "str"  [arg-type]
```

You can adjust the return type of function to be directly `str` instead of `Optional[str]` as it always return the `pod_name` and raises otherwise.

goern:
@bissenbay ping?!

bissenbay:
@fridex PTAL

bissenbay:
> Pre-Commit Test failed! Click here

@fridex I we go to this https://docs.sentry.io/platforms/python/configuration/options/, we can see that there is, indeed, `init` function in `sentry_sdk`, I really don't understand why `pre-commit` is unhappy about that?

thoth/common/logging.py
```diff
@@ -30,7 +30,7 @@
 from typing import Any
 
 from jsonformatter import JsonFormatter
-from sentry_sdk import init as sentry_sdk_init
+from sentry_sdk import init as sentry_sdk_init  # type: ignore[attr-defined]
```
fridex, line 33: Why `ignore` was needed here?

thoth/common/openshift.py
```diff
@@ -377,16 +377,38 @@ def get_pod_log(
 
         return response.text
 
-    def get_workflow_node_log(
+    def get_workflow_pod_name(
         self, node_name: str, workflow_id: str, namespace: str
-    ) -> Optional[str]:
-        """Get log from a task/node in a workflow."""
+    ) -> Optional[Any]:
```
fridex, line 382:
```suggestion
    ) -> str
```

thoth/common/openshift.py
```diff
@@ -377,16 +377,38 @@ def get_pod_log(
...
+        """Get pod name where task is being executed."""
         workflow = self.get_workflow(workflow_id, namespace=namespace)
         nodes = workflow.get("status", {}).get("nodes", {})
 
         for pod_name, node_info in nodes.items():
             if node_info["displayName"] == node_name:
-                return self.get_pod_log(pod_name, namespace=namespace, container="main")
+                return pod_name
+
+        return None
```
fridex, line 391:
```suggestion
        raise NotFoundException("No node named {node_name!r} found in workflow {workflow_id!r} in namespace {namespace!r}")
```

thoth/common/openshift.py
```diff
@@ -377,16 +377,38 @@ def get_pod_log(
...
+
+    def get_workflow_node_log(
+        self, node_name: str, workflow_id: str, namespace: str
+    ) -> Optional[str]:
+        """Get log from a task/node in a workflow."""
+        pod_name = self.get_workflow_pod_name(node_name, workflow_id, namespace)
+        if pod_name is not None:
```
fridex, line 398:
```suggestion
        return self.get_pod_log(pod_name, namespace=namespace, container="main")
```

... and you can remove the part below as the called function raises exception.

thoth/common/openshift.py
```diff
@@ -377,16 +377,38 @@ def get_pod_log(
...
+        raise NotFoundException(
+            f"No node {node_name!r} in workflow {workflow_id!r} in namespace {namespace!r} found"
+        )
+
+    def get_workflow_node_status(
+        self, node_name: str, workflow_id: str, namespace: str
+    ) -> Dict[str, Any]:
+        """Get status from a task/node in a workflow."""
+        pod_id = self.get_workflow_pod_name(node_name, workflow_id, namespace)
+        if pod_id is not None:
```
fridex, line 410:
```suggestion
    return self.get_pod_status(pod_id, namespace)
```

... and you can remove the part below as the called function raises an exception.

bissenbay:
@fridex thanks for the review
> Why ignore was needed here?
   `thoth/common/logging.py:33:1: error: Module 'sentry_sdk' has no attribute 'init'  [attr-defined]`
And I still have one more pre-commit issue `thoth/common/openshift.py:389:17: error: Returning Any from function declared to return "str"  [no-any-return]`

thoth/common/logging.py
bissenbay: When I run `pre-commit` locally, I see `thoth/common/logging.py:33:1: error: Module 'sentry_sdk' has no attribute 'init' [attr-defined]`, but looks like it is not shown when here, https://github.com/thoth-station/common/pull/1007#issuecomment-721309718

fridex:
> @fridex thanks for the review
> I still have one more pre-commit issue `thoth/common/openshift.py:389:17: error: Returning Any from function declared to return "str" [no-any-return]`

Cast to str or ignore this one. Typing does not know about types in nested objects retunred from API.

bissenbay:
> > @fridex thanks for the review
> > I still have one more pre-commit issue `thoth/common/openshift.py:389:17: error: Returning Any from function declared to return "str" [no-any-return]`
> 
> Cast to str or ignore this one. Typing does not know about types in nested objects retunred from API.

Looks like adding #type: ignore fixed, should I leave like this or cast?

fridex:
> > > @fridex thanks for the review
> > > I still have one more pre-commit issue `thoth/common/openshift.py:389:17: error: Returning Any from function declared to return "str" [no-any-return]`
> > 
> > 
> > Cast to str or ignore this one. Typing does not know about types in nested objects retunred from API.
> 
> Looks like adding `#type: ignore` fixed at https://github.com/thoth-station/common/pull/1007/files#diff-00668a241bbe4ebb4fe69571ac2634fba90bd935bfb267345ec0c946bfe6bdccR389, should I leave like this or cast?

You can keep it as is. Mind other suggestions.

thoth/common/openshift.py
```diff
@@ -377,21 +377,37 @@ def get_pod_log(
...
+            "No node named {node_name!r} found in workflow {workflow_id!r} in namespace {namespace!r}"
         )
 
+    def get_workflow_node_log(
+        self, node_name: str, workflow_id: str, namespace: str
+    ) -> Optional[str]:
+        """Get log from a task/node in a workflow."""
+        pod_name = self.get_workflow_pod_name(node_name, workflow_id, namespace)
+        return self.get_pod_log(pod_name, namespace=namespace, container="main")
+            return self.get_pod_log(pod_name, namespace=namespace, container="main")
```
fridex, line 401:
```suggestion
```

thoth/common/openshift.py
```diff
@@ -377,21 +377,37 @@ def get_pod_log(
...
+
+    def get_workflow_node_status(
+        self, node_name: str, workflow_id: str, namespace: str
+    ) -> Dict[str, Any]:
+        """Get status from a task/node in a workflow."""
+        pod_id = self.get_workflow_pod_name(node_name, workflow_id, namespace)
+    return self.get_pod_status(pod_id, namespace)
```
fridex, line 408:
```suggestion
        return self.get_pod_status(pod_id, namespace)
```

thoth/common/openshift.py
```diff
@@ -377,21 +377,37 @@ def get_pod_log(
...
+            return self.get_pod_status(pod_id, namespace)
```
fridex, line 409:
```suggestion
```

thoth/common/openshift.py
```diff
@@ -377,21 +377,35 @@ def get_pod_log(
...
         nodes = workflow.get("status", {}).get("nodes", {})
 
         for pod_name, node_info in nodes.items():
             if node_info["displayName"] == node_name:
-                return self.get_pod_log(pod_name, namespace=namespace, container="main")
+                return pod_name  # type: ignore
 
         raise NotFoundException(
-            f"No node {node_name!r} in workflow {workflow_id!r} in namespace {namespace!r} found"
+            "No node named {node_name!r} found in workflow {workflow_id!r} in namespace {namespace!r}"
```
fridex, line 392:
```suggestion
            f"No node named {node_name!r} found in workflow {workflow_id!r} in namespace {namespace!r}"
```

fridex:
I've never edited files directly in PR on GitHub. Could you please verify the code works? Otherwise, it looks like it can go in.

fridex (approved):
Tested locally, works!!!

Thanks! :+1:

PR #1010, opened 2020-10-09
Qebhwt needs deployment name

saisankargochhayat:
We need deployment name for send messages


pacospace (approved):
Thanks!! :100:

PR #1021, opened 2020-10-22
Added method to call kebechet admin workflow

saisankargochhayat:
Added calling method for Kebechet Administrator 

Related - https://github.com/thoth-station/workflow-helpers/issues/148

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1430,7 +1430,7 @@ class OpenShift:
             },
         )
 
-    def schedule_kebechet_administrator_workflow(
+    def schedule_kebechet_administrator(
         self,
         message_info: Dict[str, str],
         message_type: str,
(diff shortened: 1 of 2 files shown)


thoth/common/openshift.py
```diff
@@ -1430,6 +1430,34 @@ def schedule_kebechet_workflow(
             },
         )
 
+    def schedule_kebechet_administrator_workflow(
```
pacospace, line 1433: didn't we want to remove `_workflow` in front in the method name? we have mixed naming, in theory method in Openshift class should hide the logic, because if we don't use Argo in the future we need to rechange name to something else rather than a workflow :)

pacospace (commented):
One comment otherwise LGTM :100:

thoth/common/openshift.py
saisankargochhayat: Thanks I have changed it!

pacospace (approved):
:100: :medal_sports:

PR #1054, opened 2020-12-01
Add submit and schedule kebechet methods

xtuchyna:
https://github.com/thoth-station/mi-scheduler/pull/58

- [ ] Yes
- [X] No

New submit and schedule methods for MI inspection of Kebechet 

## Description
the `submit_mi` and `schedule_mi_workflow` methods regarding general mi analysis remained untouched

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1661,10 +1661,6 @@ class OpenShift:
         )
         return (repos, orgs)
 
-    def get_mi_kebechet_repositories(self):
-        """Get all of the repositories that use kebechet."""
-        raise NotImplementedError
-
     @staticmethod
     def parse_cpu_spec(cpu_spec: typing.Optional[str]) -> typing.Optional[float]:
         """Parse the given CPU requirement as used by OpenShift/Kubernetes."""
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1427,6 +1427,24 @@ class OpenShift:
             },
         )
 
+    def schedule_mi_kebechet_workflow(
+        self, repository: str, *, job_id: Optional[str] = None,
+    ) -> Optional[str]:
+        """Schedule Meta-information Indicators Workflow.
+
+        :param repository:str: GitHub repository in full name format: <repo_owner>/<repo_name>
+        """
+        workflow_id = job_id or self.generate_id("mi")
+        template_parameters = {"WORKFLOW_ID": workflow_id, "REPOSITORY": repository}
+
+        return self._schedule_workflow(
+            workflow=self.workflow_manager.submit_mi_kebechet,
+            parameters={
+                "template_parameters": template_parameters,
+                "workflow_parameters": {},
+            },
+        )
+
     def schedule_kebechet_workflow(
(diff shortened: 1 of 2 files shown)


fridex (approved):
:ship: :+1:

thoth/common/openshift.py
```diff
@@ -1643,6 +1661,10 @@ def get_mi_repositories_and_organizations(self,) -> Tuple[List[str], List[str]]:
         )
         return (repos, orgs)
 
+    def get_mi_kebechet_repositories(self):
```
pacospace, line 1664: This method is not taken from common, it is a query from the database that needs to be created :)

xtuchyna: yes, but cannot the query be done inside common using e.g. `psycopg2` module? (I am still trying to recreate the postgresql database locally...)

pacospace: common and storages are two different libraries and their behavior is separated, moreover, storages already depend on thoth-common, you would create a cycle in dependencies that cannot be solved. We should use libraries for their purposes. common to handle kubernetes, openshift and storages to handle database. 

The component that wants to schedule something uses common, the component the wants to query something uses storages. So in the component both libraries need to be direct dependencies :)

pacospace: > yes, but cannot the query be done inside common using e.g. `psycopg2` module? (I am still trying to recreate the postgresql database locally...)

What issues are you experiencing?

xtuchyna: > The component that wants to schedule something uses common, the component the wants to query something uses storages. So in the component both libraries need to be direct dependencies :)

yes alright, will move the method to storages then :+1:

thoth/common/openshift.py
```diff
@@ -1427,6 +1427,24 @@ def schedule_mi_workflow(
             },
         )
 
+    def schedule_mi_kebechet_workflow(
```
pacospace, line 1430: can't we extend existing method, to pass the parameter we want? I feel like mi_workflow is one and we can provide specific parameter to do something specific, WDYT?

xtuchyna: yes, that's correct. Will do that

xtuchyna: Aight, pls have a look at PR and discussion here: https://github.com/thoth-station/thoth-application/pull/692#discussion_r536031962

Kebechet workflows will be then submitted just from `mi-scheduler`

thoth/common/openshift.py
```diff
@@ -1410,14 +1410,22 @@ def schedule_qebhwt_workflow(
         )
 
     def schedule_mi_workflow(
-        self, repository: str, *, job_id: Optional[str] = None,
+        self,
+        repository: str,
+        entities: Optional[str] = None,
```
pacospace, line 1415: please adjust docstring with inputs accordingly :)

pacospace (commented):
One comment , otherwise LGTM!

thoth/common/openshift.py
xtuchyna: done!

pacospace (approved):
Thanks!!

Issue #1058, opened 2020-12-09
Consider increasing number of bits for solver document id and their names


fridex:
**Is your feature request related to a problem? Please describe.**

As of now, we use 32 bits to generate a unique identifier for solver documents:

https://github.com/thoth-station/common/blob/da1fb803bb3b131bbbf21cc393250c306f05cf81/thoth/common/openshift.py#L1052-L1054

This can lead to collisions when generating solver ids. The collision might result in overwriting documents stored on ceph.

**Describe the solution you'd like**

Consider increasing the number of bits for solver document names respecting the probability for collisions and the number of packages we want to have solved. See https://en.wikipedia.org/wiki/Birthday_attack#Mathematics as a table computing the probability of collisions.

Note we do not distinguish solver type (operating system, python version) in the solver name itself. This means that all the solvers share "namespace" in which the solver name should be generated. It might be a good idea to encode this in the solver name long term if we want to include different solvers in the system so they do not collide.

**Additional context**

As of now, we have ~2.2*10^5 package releases solved for ubi8-py36, the probability of a hash collision is more than 75% for 32 bits (see the linked wikipedia table) just for one solver. This probability grows with more solvers.

Note number of releases on PyPI (~2.2*10^6). We could include this number (and its growth) when designing the solver name.

Note also that the name of the solver is used in the workflow itself. Increasing the number of characters in the name can lead to kubernetes/openshift limitations.

Note this mechanism is used also in other parts of the project. Thus we could rething the approach in other places as well (adviser, ...).

fridex:
To avoid overwriting docs on Ceph, we could also:

 - definitely increase the number of documents we can create considering the hash collision probability
 - make sure the generated document id does not collide with an already existing one (before scheduling)
 - or... make sure the generated id does not collide with an existing one on the workflow start (after scheduling)
 - use unique constraints in the database on document id in all tables so we do not accidentally sync different document ids

goern:
sounds to me like we should go for 1 and 2. 3 and 4 seem to implement a solution at the wrong control point, if the workflow figures out that an id collides, what shall it do? fail? regenerate the id? ... ?

fridex:
> 3 and 4 seem to implement a solution at the wrong control point, if the workflow figures out that an id collides, what shall it do? fail? regenerate the id? ... ?

[...]

PR #1069, opened 2021-01-06
Increase random bits used to distinguish workflows

fridex:

Fixes: #1058 

- [x] No

With this change, we will increase the number of bits needed to distinguish documents. 128. With this, we should be able to analyze 2.6×10^10 packages per solver with a probability of  10^(-18) for collision. Kubernetes pod name can hold up to 253 characters, this digest requires 32 characters, argo adds 11 characters to distinguish pods within workflow, cca ~20 is required for solver prefix and the rest is required for delimiters and we still have some free space.

https://en.wikipedia.org/wiki/Birthday_problem#Probability_table

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1042,12 +1042,12 @@ class OpenShift:
         # name and failed with regexp issues (that were not related to the
         # generateName configuration).
         if prefix and identifier:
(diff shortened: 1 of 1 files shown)


harshad16 (approved):
lgtm, thanks for the update.

PR #1093, opened 2021-01-20
Revisit generating id to avoid hash collisions

fridex:

With this change, we add datetime to the suffix generated. It will provide the capability to distinguish when the given document was generated just by looking at the document identifier. Another, more important aspect, is that the timestamp encoded will avoid any collisions considering additional suffix. It means that the hash generation in the exact same second would return the same 64 UUID hash, which probability is very close to zero given the random seed.

- [x] No

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -18,6 +18,7 @@
 """Handling OpenShift and Kubernetes objects across project."""
 
 import os
+import datetime
 import logging
 import requests
 import typing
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -1074,17 +1075,14 @@ def generate_id(
         prefix: Optional[str] = None, identifier: Optional[str] = None
     ) -> str:
         """Generate an identifier."""
-        # A very first method used 'generatedName' in ImageStream configuration,
-        # but it looks like there is a bug in OpenShift as it did not generated any
-        # name and failed with regexp issues (that were not related to the
-        # generateName configuration).
+        suffix = f"-{datetime.datetime.now():%y%m%d%H%M%S}-{random.getrandbits(64):08x}"
```
KPostOffice, line 1078: No "-" required at the beginning of the format string as it is added later.

KPostOffice (commented):
One small comment

thoth/common/openshift.py
fridex: Thanks for the review, nice catch :+1:

Added at the end to keep function call compatible with the previous implementation. Thanks!

KPostOffice (approved):
LGTM :+1:

fridex:
> LGTM

Thanks!
/approve

PR #1098, opened 2021-01-25
add enum for internal triggers

KPostOffice:

thoth-station/kebechet#584

- [ ] Yes
- [x] No

Create an enum representing the different internal triggers

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,12 @@
 ---
 repos:
   - repo: git://github.com/Lucas-C/pre-commit-hooks
-    rev: v1.1.1
+    rev: v1.1.9
     hooks:
       - id: remove-tabs
 
   - repo: git://github.com/pre-commit/pre-commit-hooks
-    rev: v2.0.0
+    rev: v3.4.0
     hooks:
       - id: trailing-whitespace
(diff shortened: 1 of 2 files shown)


fridex (commented):
Looks good to me :+1: :1st_place_medal:

KPostOffice:
> Looks good to me +1 1st_place_medal

Should I add one more line:
`NONE = auto()`
or in the case of no justification should we just use python's `None`?

pacospace:
> > Looks good to me +1 1st_place_medal
> 
> Should I add one more line:
> `NONE = auto()`
> or in the case of no justification should we just use python's `None`?

is there a case where the internal trigger run without a justification?

KPostOffice:
I don't think internal triggers should be sent without this justification, but this is metadata that will be part of `kebechet-administrator-trigger` message.  At that point anything could technically send a `kebechet-adminstrator` message I guess? though I feel like `kebechet-admin` will probably only be used for multiplexing internal triggers to schedule Kebechet for each repo so it should always have this metadata set.  I will leave this as is.

fridex (approved):
lgtm, thanks! :100:

harshad16 (approved):
lgtm :+1:

PR #1100, opened 2021-02-02
Fix package-extract job id propagation

fridex:

Fixes: https://github.com/thoth-station/thoth-application/issues/848

- [x] No

When `job_id` is not supplied, THOTH_DOCUMENT_ID is an empty string.


harshad16 (commented):
/lgtm

PR #1103, opened 2021-02-04
add metadata to relevant schedule methods

KPostOffice:

thoth-station/kebechet#584

- [ ] Yes
- [x] No

Add metadata to both kebechet-run-url which is triggered by `kebechet-admin` in workflow helpers as well as `adviser` so that the metadata can be passed when using advise and provenance checkers.  The metadata will be used by kebechet to construct PR and issue bodies. In the case of the advise and provenance manager, the metadata will be sent to adviser and provenance workflows and then be received by kebechet-run-results.

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1341,6 +1341,7 @@ class OpenShift:
         github_base_repo_url: Optional[str] = None,
         re_run_adviser_id: Optional[str] = None,
         source_type: Optional[str] = None,
+        kebechet_metadata: Optional[Dict[str, Any]] = None,
     ) -> Optional[str]:
         """Schedule an adviser run."""
         if not self.backend_namespace:
@@ -1397,6 +1398,7 @@ class OpenShift:
                 "origin": origin,
                 "re_run_adviser_id": re_run_adviser_id,
                 "source_type": source_type if source_type is not None else None,
+                "kebechet_metadata": kebechet_metadata,
             }
         )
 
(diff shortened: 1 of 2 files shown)


pacospace (commented):
only one question, otherwise lgtm!

thoth/common/openshift.py
```diff
@@ -1588,6 +1593,7 @@ def schedule_kebechet_run_url_workflow(
         repo_url: str,
         service_name: str = "github",
         *,
+        kebechet_metadata: Dict[str, Any] = {},
```
pacospace, line 1596: why not `Optional[Dict[str, Any]] = None`?

thoth/common/openshift.py
```diff
@@ -1452,7 +1455,9 @@ def schedule_provenance_checker(
         template_parameters = {
             "THOTH_ADVISER_REQUIREMENTS": requirements,
             "THOTH_ADVISER_REQUIREMENTS_LOCKED": requirements_locked,
-            "THOTH_ADVISER_METADATA": json.dumps({"origin": origin}),
+            "THOTH_ADVISER_METADATA": json.dumps(
+                {"origin": origin, "kebechet_metadata": kebechet_metadata}
```
fridex, line 1459: We should sync on this one. I opened https://github.com/thoth-station/adviser/pull/1625 for this and the semantics behind that is:

Keep metadata generic, so anything can be passed in the JSON. Any specific entries like `stack_info` or justification propagated to the adviser output should go to `thoth.adviser` key so we can possibly expand the metadata. WDYT?

KPostOffice: I'm not sure that I understand what you mean.

KPostOffice: @fridex Are you saying get rid of https://github.com/thoth-station/common/pull/1103/files#diff-00668a241bbe4ebb4fe69571ac2634fba90bd935bfb267345ec0c946bfe6bdccL1338-L1343 and put it all under the generic argument `metadata` so we can easily add new metadata?

fridex (approved):
Thanks :+1:  I will open a follow-up PR to review/sync.

fridex:
@KPostOffice PTAL on conflicts.

PR #1107, opened 2021-02-08
Thoth application#398

KPostOffice:

https://github.com/thoth-station/thoth-application/pull/786/files

- [ ] Yes
- [x] No

add params for sending message and whether graph-sync should force sync

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -833,8 +833,6 @@ class OpenShift:
         *,
         raw_specification: Optional[Dict[str, Any]] = None,
         job_id: Optional[str] = None,
-        send_messages: Optional[int] = None,
-        force_sync: Optional[int] = None,
     ) -> Optional[str]:
         """Schedule an inspection run."""
         if not self.amun_inspection_namespace:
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -858,6 +860,12 @@ def schedule_inspection(
             "memory"
         ]
 
+        if send_messages:
+            template_parameters["SEND_MESSAGES"] = str(send_messages)
+
+        if force_sync:
+            template_parameters["FORCE_SYNC"] = str(force_sync)
```
fridex, line 867: Can we prefix these with `THOTH_`?

KPostOffice: Yeah, I'll change them here and in the linked PR

thoth/common/openshift.py
```diff
@@ -833,6 +833,8 @@ def schedule_inspection(
         *,
         raw_specification: Optional[Dict[str, Any]] = None,
         job_id: Optional[str] = None,
+        send_messages: Optional[int] = None,
+        force_sync: Optional[int] = None,
```
fridex, line 837: Can these become boolean? It sounds to me like they have just binary values.

KPostOffice: @fridex I removed these two from args to the function. Within `specification` they are stored as bools and then cast `str(int(foo))` to the template params so that they end up as `1` or `0`

fridex (approved):
lgtm :+1:

PR #1113, opened 2021-02-18
Fix obtaining pod status for workflows

fridex:

Fixes: https://github.com/thoth-station/integration-tests/issues/128

- [x] No

## Description

User API schema validation fails as obtaining status of a node in argo workflow does not translate response to the required schema - rather it returns directly Kubernetes low lever pod info which is undesired.


harshad16 (commented):
/lgtm :1234:

PR #1141, opened 2021-03-23
Add additional configuration as parameter

xtuchyna:
Related to https://github.com/thoth-station/thoth-application/pull/1098#event-4497551515

- [ ] Yes
- [x] No

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1505,6 +1505,7 @@ class OpenShift:
         repository: str,
         entities: Optional[str] = None,
         knowledge_path: Optional[str] = None,
+        additional_configuration: Optional[str] = None,
         *,
         job_id: Optional[str] = None,
     ) -> Optional[str]:
@@ -1520,6 +1521,7 @@ class OpenShift:
             "REPOSITORY": repository,
             "ENTITIES": entities,
             "KNOWLEDGE_PATH": knowledge_path,
+            "ADDITIONAL_CONFIGURATION": additional_configuration,
         }
 
         return self._schedule_workflow(
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1505,7 +1505,7 @@ class OpenShift:
         repository: str,
         entities: Optional[str] = None,
         knowledge_path: Optional[str] = None,
-        mi_thoth: bool = False,
+        mi_used_for_thoth: bool = False,
         *,
         job_id: Optional[str] = None,
     ) -> Optional[str]:
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1521,7 +1521,7 @@ class OpenShift:
             "REPOSITORY": repository,
             "ENTITIES": entities,
             "KNOWLEDGE_PATH": knowledge_path,
-            "MI_THOTH": "true" if mi_thoth else "false",
+            "MI_THOTH": "true" if mi_used_for_thoth else "false",
         }
 
         return self._schedule_workflow(
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1505,7 +1505,7 @@ class OpenShift:
         repository: str,
         entities: Optional[str] = None,
         knowledge_path: Optional[str] = None,
-        additional_configuration: Optional[str] = None,
+        mi_thoth: bool = False,
         *,
         job_id: Optional[str] = None,
     ) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -1505,6 +1505,7 @@ def schedule_mi_workflow(
         repository: str,
         entities: Optional[str] = None,
         knowledge_path: Optional[str] = None,
+        additional_configuration: Optional[str] = None,
```
fridex, line 1508: Are you going to pass parameters such as `-m` to the container?

xtuchyna: yes, e.g. `"-m"` or `"-mt"` or `"-t"`, possibly few others in future

xtuchyna: I think this is a good idea, wdyt @fridex ?

pacospace: maybe there should be an enum or something that prevent from passing any string there in this case

pacospace: Or convert those string in parameters that are booleans that activate, deactivate something, so having specific environment variables for each of those letters

xtuchyna: This was meant just as a quick patch for the other parameters passing. Didn't want to add every one of them, because we still don't know how many of them will be used. Currently there are flags like ` process_knowledge`, `visualize_statistics`, `reviewer_reccomender`, `thoth` and `metrics` and some of them can be removed in future, or potentially changed.

The purpose was to just use passing by string now and in near future change that to explicit parameters. I can make TODO Issues out of this so that it does not get lost through time.

This additional parameters validaiton can be also done in `mi-scheduler` in the end, as the repositories and organizations are also checked there.

pacospace: isn't it enough to have the string assigned in the template for now then? https://github.com/thoth-station/thoth-application/pull/1098#event-4497551515

xtuchyna: there are basically two workflow types now, the kebechet one and the general mi.
Kebechet uses `"-t"` and the general one will target the `"-m"` option, so therefore I didn't specify anything in the template

xtuchyna: so what's the status of this @fridex @pacospace ?

pacospace: let's add a TODO if this is stopping you for now and let's plan those adjustments. @fridex thoughts?

xtuchyna: I've referenced this as a future TODO in `mi-scheduler` here: https://github.com/thoth-station/mi-scheduler/issues/109
therefore marking this as resolved for now

fridex: I'm not very happy with merging this as it introduces a tech dept. The flags can be mapped to environment variables (with [click, it's just one line](https://click.palletsprojects.com/en/7.x/options/#values-from-environment-variables)) and mapped to parameters supplied to the method on an `OpenShift` instance. Otherwise, the proposed solution is hard to maintain, hard to track the request and hard to extend the solution if the call corresponds to just `openshift.schedule_mi_workflow(additional_paramters="-mt")`. It's prone to errors and does not fit to the rest of the system.

If there is time pressure, let's merge this. But putting more of these things into the application can boil to tech dept we will spend weeks to reduce (I've already been there and it's not something I would like to repeat).

xtuchyna: alright, parameter changed to be more explicit!

thoth/common/openshift.py
```diff
@@ -1520,6 +1521,7 @@ def schedule_mi_workflow(
             "REPOSITORY": repository,
             "ENTITIES": entities,
             "KNOWLEDGE_PATH": knowledge_path,
+            "MI_THOTH": "true" if mi_thoth else "false",
```
pacospace, line 1524: why this booleans are strings?

thoth/common/openshift.py
```diff
@@ -1505,6 +1505,7 @@ def schedule_mi_workflow(
         repository: str,
         entities: Optional[str] = None,
         knowledge_path: Optional[str] = None,
+        mi_thoth: bool = False,
```
pacospace, line 1508: what is this variable used for?  do you mean like `mi_used_for_thoth`?

thoth/common/openshift.py
xtuchyna: it's a specified parameter that will be used as envvar for setting the cli option `-t`, and the envvar are always strings (e.g. see https://github.com/thoth-station/common/blob/5b96cd3c2ddf4856426f69f87421215643f3f4cc/thoth/common/openshift.py#L1368 - setting integer, however in string form, because of envvar)
I don't know whether `click` supports assigning "1" or "0" as a parameter, but I know that it supports strings "false" and "true" when it comes to setting flags by envvars

thoth/common/openshift.py
xtuchyna: true, commited your suggestion

pacospace (approved):
lgtm!

PR #1158, opened 2021-05-05
Run purge workflows in middletier namespace

fridex:

We do not have argo workflows in graph namespace, let's utilize middletier namespace. Thanks to @harshad16 for spotting this issue.

- [x] No


harshad16 (approved):
lgtm, thanks :100:

PR #1171, opened 2021-06-14
update advises

KPostOffice:

thoth-station/thoth-application#1655

- [ ] Yes
- [x] No

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1348,6 +1348,7 @@ class OpenShift:
         kebechet_metadata: Optional[Dict[str, Any]] = None,
         justification: Optional[List[Dict[str, Any]]] = None,
         stack_info: Optional[List[Dict[str, Any]]] = None,
+        package_update_info: Optional[Dict[str, str]] = None,
     ) -> Optional[str]:
         """Schedule an adviser run."""
         if not self.backend_namespace:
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -1404,6 +1405,9 @@ def schedule_adviser(
         if count is not None:
             template_parameters["THOTH_ADVISER_COUNT"] = str(count)
 
+        if from_package_update:
+            template_parameters["THOTH_ADVISER_DEPLOYMENT_PACKAGE_UPDATE"] = "1"
```
fridex, line 1409: Any reason not to put directly `THOTH_ADVISER_DEPLOYMENT_PACKAGE_UPDATE` - JSON describing the package to be updated?

KPostOffice: ++, I realized this information will be present in Kebechet metadata. Should we just duplicate it here if the message type is package_release?

fridex: Yes, I think so - adviser will need to be triggered with adjusted environment variables to run in the "update" mode. The function can accept an optional triplet (package name, package version, index url) that could be serialized to the env variable. The triplet can be supplied based on the message received.

goern:
Can we merge this?

fridex (approved):
lgtm, thanks 👍🏻

PR #1183, opened 2021-07-02
get update info from kebechet meta

KPostOffice:

Passing package_update_info around is unnecessary. The information required is present in `kebechet_metadata`

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -48,7 +48,7 @@ from .helpers import (
     normalize_os_version,
     format_datetime,
 )
-from .enums import ThothAdviserIntegrationEnum
+from .enums import ThothAdviserIntegrationEnum, InternalTriggerEnum
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


thoth/common/openshift.py
```diff
@@ -1315,7 +1315,6 @@ def schedule_adviser(
         kebechet_metadata: Optional[Dict[str, Any]] = None,
         justification: Optional[List[Dict[str, Any]]] = None,
         stack_info: Optional[List[Dict[str, Any]]] = None,
-        package_update_info: Optional[Dict[str, str]] = None,
```
pacospace, line 1318: what component was using this parameter?

pacospace (commented):
One comment, otherwise lgtm!

thoth/common/openshift.py
KPostOffice: https://github.com/thoth-station/user-api/pull/1456

Was never actually pulled in

pacospace: Thanks, then I think we are good here :)

pacospace (approved):
Thanks 💯

PR #1187, opened 2021-07-08
Add split for spaces

xtuchyna:
the `def get_mi_repositories_and_organizations` was giving back names with spaces around them (id did not strip them from https://github.com/thoth-station/thoth-application/blob/7f48f685731db12fc385a3d8d3bda6437e77b0d6/mi-scheduler/base/configmaps.yaml#L9)
This fixes that

- [ ] Yes
- [x] No

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1709,7 +1709,7 @@ class OpenShift:
             if repositories is not None and repositories != ""
             else []
         )
-        return (repos, orgs)
+        return ([r.split() for r in repos], [o.split() for o in orgs])
 
     @staticmethod
     def parse_cpu_spec(cpu_spec: typing.Optional[str]) -> typing.Optional[float]:


pacospace:
I think it is faster adjusting the "thoth-station, AICoE, aicoe-aiops" -> to "thoth-station,AICoE,aicoe-aiops" in thoth-application?

thoth/common/openshift.py
```diff
@@ -1709,7 +1709,7 @@ def get_mi_repositories_and_organizations(
             if repositories is not None and repositories != ""
             else []
         )
-        return (repos, orgs)
+        return ([r.split() for r in repos], [o.split() for o in orgs])
```
harshad16, line 1712: is the output seperate with commas "," ? sorry i don't remember how the repos were structured.
this would split with spaces , is that what we want ?

harshad16 (commented):
just a minor question

thoth/common/openshift.py
xtuchyna: yes exactly

xtuchyna: names are divided by commas, however there can be optional number of spaces in between the name and comma

xtuchyna:
Based on https://github.com/thoth-station/common/pull/1187#issuecomment-876550568, I am closing the PR, and instead changing the string list to format wo spaces

PR #1193, opened 2021-07-29
Add merge path to schedule mi workflows

xtuchyna:
Related to https://github.com/thoth-station/mi-scheduler/pull/190

- [ ] Yes
- [X] No

Add new merge path variable to the scheduling of mi workflows

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1490,7 +1490,7 @@ class OpenShift:
             "KNOWLEDGE_PATH": knowledge_path,
             "MI_THOTH": "1" if mi_used_for_thoth else "0",
             "MI_MERGE": "1" if mi_merge else "0",
-            "MI_MERGE_PATH": mi_merge_path,
+            "MI_MERGE_PATH": mi_merge_path or "",
         }
 
         return self._schedule_workflow(
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -1472,6 +1472,7 @@ class OpenShift:
         knowledge_path: Optional[str] = None,
         mi_used_for_thoth: Optional[bool] = False,
         mi_merge: Optional[bool] = False,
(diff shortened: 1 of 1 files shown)


xtuchyna:
Also related to https://github.com/thoth-station/mi/pull/439

thoth/common/openshift.py
```diff
@@ -1472,6 +1472,7 @@ def schedule_mi_workflow(
         knowledge_path: Optional[str] = None,
         mi_used_for_thoth: Optional[bool] = False,
         mi_merge: Optional[bool] = False,
+        mi_merge_path: Optional[str] = None,
```
pacospace, line 1475: it can be None?

xtuchyna: yes, if `mi_merge = True`  then the default location will be used.
If it is `False`, then we don't care and it is None.

thoth/common/openshift.py
```diff
@@ -1489,6 +1490,7 @@ def schedule_mi_workflow(
             "KNOWLEDGE_PATH": knowledge_path,
             "MI_THOTH": "1" if mi_used_for_thoth else "0",
             "MI_MERGE": "1" if mi_merge else "0",
+            "MI_MERGE_PATH": mi_merge_path,
```
fridex, line 1493:
```suggestion
            "MI_MERGE_PATH": mi_merge_path or "",
```

I think we assign an empty string if a parameter is `None`. Is that OK here (I don't have mi context here)?

xtuchyna: Yes, it is okay

thoth/common/openshift.py
xtuchyna: @pacospace wdyt?

fridex (approved):
lgtm, thanks! 👍🏻

Issue #1219, opened 2022-01-19
Replace argo-workflows with hera


fridex:
**Is your feature request related to a problem? Please describe.**

See https://www.dynotx.com/hera-the-missing-argo-workflows-python-sdk/

We should stop using argo-workflows that are now archived and use hera instead.

PR #1223, opened 2022-02-07
Update workflows.py

peremoar:
From v4.0.0, [&lowbar;&lowbar;init&lowbar;&lowbar;.py](https://github.com/argoproj-labs/argo-client-python/tree/master/argo/workflows) file of argo-client-python is empty. For this reason, the import should be now:

`from argo.workflows.client import models
`
instead of:

`from argo.workflows import models
`
Other versions of the client (as [this one](https://github.com/CermakM/argo-client-python/tree/master/argo/workflows)) keeps the older content of the  &lowbar;&lowbar;init&lowbar;&lowbar;.py.

The change doesn't produces backward incompatibilities to users with the non-empty version of the &lowbar;&lowbar;init&lowbar;&lowbar;.py

…

- [ ] Yes
- [x] No

… Explain your changes.

### DIFF
--- a/thoth/common/workflows.py
+++ b/thoth/common/workflows.py
@@ -31,7 +31,7 @@ from typing import Optional
 from typing import Union
 
 from argo.workflows import client
-from argo.workflows import models
+from argo.workflows.client import models
 
 from .exceptions import ConfigurationError
 from .exceptions import WorkflowError


fridex:
Thanks, @peremoar for your pull-request. We [pin argo-workflows to `<4`](https://github.com/thoth-station/common/blob/13e47c8df393f9fe68a8827b02fcd63e207f0411/Pipfile#L8) and there are no plans to update to a more recent version. We would like to migrate to hera-workflows (see https://github.com/thoth-station/common/issues/1219), which is more likely an alternative and update to the codebase.

Thanks again for your pull request and your interest. I will close it as the codebase will require a new library.

/close

PR #1231, opened 2022-03-08
Send messages based on template defaults, do not hardcode

fridex:

We should not hardcode this behavior, rather have it configurable on the template level.

### DIFF
--- a/thoth/common/openshift.py
+++ b/thoth/common/openshift.py
@@ -864,12 +864,15 @@ class OpenShift:
             "memory"
         ]
 
-        template_parameters["THOTH_SEND_MESSAGES"] = str(
-            int(specification.get("send_messages", True))
-        )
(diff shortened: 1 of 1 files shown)


harshad16 (approved):
lgtm
thanks :100:

thoth/common/openshift.py
```diff
@@ -864,9 +864,11 @@ def schedule_inspection(
...
 
-        template_parameters["THOTH_SEND_MESSAGES"] = str(
-            int(specification.get("send_messages", True))
-        )
+        if specification.get("send_messages") is not None:
+            template_parameters["THOTH_SEND_MESSAGES"] = str(
+                int(specification["send_messages"])
+            )
+
         template_parameters["THOTH_FORCE_SYNC"] = str(
```
KPostOffice, line 872: The same should be done here.

KPostOffice (commented):
/approve

PR #1250, opened 2022-04-06
Autoupdate pre-commit, update prow image tags

xtuchyna:
## Related Issues and Dependencies
None

## This introduces a breaking change

- [ ] Yes
- [X] No

## Description
SSIA

PR #1253, opened 2022-04-11
fix pre-commit by removing mypy 

harshad16:

#1250 

## Description

Testing done
we have removed the mypy check for now, will be included in a separate PR.


fridex:
lgtm, merging as discussed

Issue #1251, opened 2022-04-07
kubernetes module version 11.0.0 has CVE


goern:
**Is your feature request related to a problem? Please describe.**
https://github.com/thoth-station/common/blob/master/requirements.txt#L10 declares a dependency on kubernetes, current versions v11.0.0 is affected by https://access.redhat.com/security/cve/cve-2020-1747

This results in all our container images having a critical security issue, for example: https://quay.io/repository/thoth-station/integration-tests/manifest/sha256:7d4195a824800c12b63b011e29f2f1ffc7e623a3753373cd5c2c5c775b3ac7df?tab=vulnerabilities

**Describe the solution you'd like**
update kubernetes dependency to new version, release new patch version of thoth-common

**Describe alternatives you've considered**
n/a 

**Additional context**

Update to v12.0.0 can be done,
due to the issue https://github.com/thoth-station/common/issues/1273
wait for a better release and testing.

harshad16:
Changing the priority as upgrading to v12.0.0 kubernetes would cause failure in user-api.
/priority important-soon
/lifecycle frozen
/triage accepted


Any upgrade of kubernetes and openshift version, should be tested in user-api as well.
- Install common in user-api
- Execute the user-api to check its functionalities.

codificat:
I think this is more:
/remove-kind feature
/kind bug

PR #1254, opened 2022-04-11
bump kubernetes version for common image

Gregory-Pereira:

Fixes: https://github.com/thoth-station/common/issues/1251

- [ ] Yes
- [x] No

Bumping version of PyYaml (which depends on Kubernetes, which depends on openshift) for CVE_2020_1747

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -8,10 +8,10 @@ argo-workflows = ">=3.5,<4"
 daiquiri = "*"
 requests = "*"
 sentry-sdk = "*"
-kubernetes = "~=11.0.0"  # required pyyaml>=3.12
-openshift = "==0.11.0"
+kubernetes = ">=12.0.1"  # required pyyaml>=5.3.1
+openshift = "~=0.12.0"
 attrs = "*"
-pyyaml = ">=3.12"
+pyyaml = ">=5.3.1"
 flexmock = "*"
 jsonformatter = "*"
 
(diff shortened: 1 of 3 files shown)


harshad16 (changes requested):
please revert pre-commit chnages

Pipfile
```diff
@@ -8,10 +8,10 @@ argo-workflows = ">=3.5,<4"
 daiquiri = "*"
 requests = "*"
 sentry-sdk = "*"
-kubernetes = "~=11.0.0"  # required pyyaml>=3.12
-openshift = "==0.11.0"
+kubernetes = ">=12.0.1"  # required pyyaml>=5.3.1
+openshift = "==0.12.0"
```
harshad16, line 12:
```suggestion
openshift = "~=0.12.0"
```
How about this https://github.com/openshift/openshift-restclient-python/tree/v0.12.1
It has some changes: https://github.com/openshift/openshift-restclient-python/commits/v0.12.1

harshad16 (changes requested):
All lgtm
can you please rebase your PR that would get to 1 commit 
and also one suggestion.

fridex (approved):
Changes look good to me. I've also tried to verify that the common library is still able to schedule workfload in the cluster. No issues observed, but we should keep an eye on this change. 👍🏻

